PR #350, opened 2026-05-08
🐛 fix(i18n): align Crowdin locale exports

scttbnsn:
## Summary
- map Crowdin language codes to the locale folder IDs supported by the app
- limit Crowdin downloads to the languages currently exposed in the locale picker
- add a guard test so future Crowdin config drift is caught before generated sync PRs add ignored locale folders

## Verification
- node ./node_modules/vitest/vitest.mjs run tests/boot/crowdin-config.spec.ts --maxWorkers=1 --fileParallelism=false
- /Users/sbenson/code/drydock/node_modules/.bin/biome check .github/workflows/i18n-crowdin.yml crowdin.yml ui/tests/boot/crowdin-config.spec.ts

Supersedes generated Crowdin sync PR #348, which exported unsupported region-coded locale folders.

### DIFF
--- a/.github/workflows/i18n-crowdin.yml
+++ b/.github/workflows/i18n-crowdin.yml
@@ -31,6 +31,17 @@ jobs:
           upload_sources: true
           upload_translations: true
           download_translations: true
+          download_translations_args: >-
+            --language=de
+            --language=es-ES
+            --language=fr
+            --language=it
+            --language=nl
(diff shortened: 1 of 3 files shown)


scttbnsn:
Superseded by #351 so the rc18 dependency gate fix and Crowdin locale export fix ship through one release branch.

Issue #363, opened 2026-05-14
1.5.0-rc.20 drydock in a startup crash loop


begunfx:
### Drydock version

1.5.0-rc.20

### What happened?

Seeing the following in the drydock container log.  Seems like something to do with how the secrets are being shared between agents and drydock.

```
file:///home/node/app/dist/agent/AgentClient.js:103

            throw new Error(`Agent ${this.name} is configured with a secret over insecure HTTP (${this.baseUrl}). Configure HTTPS (certfile/cafile) to protect X-Dd-Agent-Secret.`);

                  ^

Error: Agent servicevault2 is configured with a secret over insecure HTTP (http://servicevault2:3010). Configure HTTPS (certfile/cafile) to protect X-Dd-Agent-Secret.

    at AgentClient.rejectSecretConfiguredOverHttp (file:///home/node/app/dist/agent/AgentClient.js:103:19)

    at new AgentClient (file:///home/node/app/dist/agent/AgentClient.js:62:14)

    at file:///home/node/app/dist/agent/index.js:17:24

    at Array.forEach (<anonymous>)

    at Module.init (file:///home/node/app/dist/agent/index.js:9:25)

    at file:///home/node/app/dist/index.js:50:28

    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)

Node.js v24.14.0
```

### What did you expect?

See above

### Steps to reproduce

See above

### Relevant configuration

```yaml
Drydock compose stack:

services:
  drydock:
    image: ghcr.io/codeswhat/drydock:1.5.0-rc.20
    container_name: drydock
    restart: unless-stopped
    ports:
      - "3010:3000"
    depends_on:
      socket-proxy:
        condition: service_healthy
    volumes:
      - /volume1/docker/drydock:/store
    labels:
      - "dd.watch=true"
    environment:
      # --- Oauth2/OpenID ---
      - DD_PUBLIC_URL=<reacted>
      - DD_AUTH_OIDC_AUTHENTIK_CLIENTID=<redacted>
      - DD_AUTH_OIDC_AUTHENTIK_CLIENTSECRET=<redacted>
      - DD_AUTH_OIDC_AUTHENTIK_DISCOVERY=<redacted>
      - DD_AUTH_OIDC_AUTHENTIK_INSECURE=true #for self-signed certificates use this to skip TLS verification for OIDC
      - DD_AUTH_OIDC_AUTHENTIK_REDIRECT=false # optional (to skip internal login page)
      
[...]

Issue #362, opened 2026-05-14
Agents displaying 0 running containers


Cesc1986:
### Drydock version

v1.5.0-rc.19

### What happened?

Hi,

again, great piece of software. May be due to the steep learning curve and complex setup an error on my behalf, but I see constantly single or multiple agents reporting 0 running containers, even though there are runnig containers on that machine. I couldn't detect a pattern, its sometimes agent 1, sometimes agent 2, sometimes agent 3, sometimes all are working. All agents have exactly the same configuration, password, PATs, etc.

Take this case, only my host "apps" is reporting 0 containers running. 

<img width="1192" height="240" alt="Image" src="https://github.com/user-attachments/assets/6acaf937-8769-4d16-adc6-203498c46424" />

This is strange, as the actual drydock agent running on this machine is detecting 25 containers. When I restart the local host machine, I think I have observed that all 3 agents are reporting the correct amount of containers as running.

````
[08:00:49.187] INFO (drydock/7): Cron started (0 * * * *) {"component":"watcher.docker.local"}

[08:00:49.222] WARN (drydock/7): immich_postgres: Cannot get a reliable tag for this image [ghcr.io/immich-app/postgres:14-vectorchord0.4.3-pgvectors0.2.0@sha256:bcf63357191b76a916ae5eb93464d65c07511da41e3bf7a8416db519b40b1c23] {"component":"watcher.docker.local"}

[08:00:49.228] WARN (drydock/7): immich_redis: Cannot get a reliable tag for this image [docker.io/valkey/valkey:9@sha256:3b55fbaa0cd93cf0d9d961f405e4dfcc70efe325e2d84da207a0a8e6d8fde4f9] {"component":"watcher.docker.local"}

[08:00:49.230] WARN (drydock/7): local_immich_redis - No Registry Provider found

[08:00:49.617] WARN (drydock/7): Error when processing (Request failed with status code 429) {"component":"watcher.docker.local","container":"local_immich_postgres"}

[08:00:54.082] WARN (drydock/7): Error when processing (Request failed with status code 429) {"component":"watcher.docker.local","container":"local_immich_machine_learning"}

[08:00:57.269] WARN (drydock/7): GitHub release notes lookup is rate-limited {"component":"release-notes.provider.github"}

[08:01:00.814] WARN (drydock/7): GitHub release notes lookup is rate-limited {"component":"release-notes.provider.github"}

[...]

PR #364, opened 2026-05-14
Release v1.5.0-rc.21 prep

scttbnsn:
## Summary

Cutting v1.5.0-rc.21. Small surface: four user-visible changes plus the polish + hygiene work needed to land them cleanly.

### Fixes
- 🐛 **#362, #363 — `DD_SESSION_SECRET` no longer crashes startup when unset.** rc.20 made the env var a hard requirement to stop the per-process random fallback from invalidating sessions on every restart; the fallback is now restored as a *persisted* secret in a new `secrets` store collection (env override > stored > generate-and-persist). Existing deployments upgrade without setting anything; sessions survive restarts.
- 🐛 **discussion #295 — Release-notes icon now opens the same popover regardless of metadata shape.** Containers with only a bare `releaseLink` URL used to render a direct external anchor (no popover); they now render the same popover shell with a single external-link row, matching the structured-notes case.

### Features
- ✨ **`DD_AGENT_ALLOW_INSECURE_SECRET` escape hatch for closed-LAN agents (#362 issue 1, #363).** rc.20 tightened the agent-secret-over-HTTP check from `warn` to `throw`; this flag downgrades it back to `warn` for operators who explicitly accept that the agent secret travels in cleartext on an isolated LAN. Default behavior is unchanged — without the flag the throw still fires. Any value other than exact `true` still throws.

### Performance
- ⚡ **Default watcher cron relaxed from hourly to every 6 hours (#342 follow-up).** Hourly was the most aggressive default among active 2026 update managers (Diun ships `0 */6 * * *`, archived Watchtower defaulted to 24h). With fleets of 20+ containers and image tag lists that paginate to thousands of entries, hourly saturated anonymous Docker Hub limits and GitHub's release-notes ceiling. rc.20's per-host token bucket stays as the safety net; this addresses the root cause. Operators who set `DD_WATCHER_{name}_CRON` explicitly are unaffected.

### Tests / hygiene
- 🧪 New `scripts/bench/` regression baselines: `registry-rate` (validates the cron default change), `aggregator-tick` (rc.17 stats aggregator), `lock-manager` (rc.17 concurrency primitives). Three historical `bench-301-*` scripts renamed to topic-based filenames.
- 🧪 Coverage gate restored to 100% (`./secrets` mock added to `store/index.test.ts`; new popover guards marked `c8 ignore` for jsdom-unreachable paths; `ContainersGroupedViews` tagName assertion updated for the new popover trigger).
- 🎨 Bench scripts biome-formatted; bench docs satisfy markdownlint MD040 / MD022.
- 🔧 Pre-push coverage timeout bumped 6m → 10m so the gate no longer trips intermittently with `signal: killed` on green coverage.

## Test plan
- [ ] CI Verify passes on this PR
- [ ] Manual: confirm `DD_AGENT_ALLOW_INSECURE_SECRET=true` downgrades the rc.20 throw to a startup warn (and absent flag still throws)
- [ ] Manual: confirm fresh boot with no `DD_SESSION_SECRET` set generates and persists a value to `/store/dd.json`, and sessions survive a restart
- [ ] Manual: hover the release-notes icon for a container with only `releaseLink` and confirm the popover opens with an external-link row
- [ ] After merge: dispatch `release-cut.yml` from main with `release_tag=v1.5.0-rc.21`


biggest-littlest (approved):
Approving rc.21 prep (code-owner review).

ALARGECOMPANY (approved):
Approving rc.21 prep (code-owner review).

PR #366, opened 2026-05-15
Release v1.5.0-rc.21 tail: i18n + Dockerfile arch-drift fix

scttbnsn:
## Summary

Tail commits for rc.21. The rc.21 prep PR (#364) merged cleanly, but the release-cut workflow failed at the post-merge CI Verify because the Docker build couldn't resolve a stale Alpine package pin on `linux/arm64`. Bundling that fix with the four i18n extractions discussion #329 was waiting on.

After this lands, dispatch `release-cut.yml` again with `release_tag=v1.5.0-rc.21`. No tag exists yet, so rc.21 wraps the new HEAD.

### Fixed
- 🐛 **Docker multi-arch build no longer fails when Alpine repos drift between archs.** `Dockerfile` pinned `curl=8.17.0-r1`, which broke `linux/arm64` after Alpine's `latest-stable/aarch64` mirror rotated to `curl-8.19.0-r0` while `latest-stable/x86_64` was still on `8.17.0-r1`. No single pin can satisfy both archs during a mirror rotation window; the curl entry is now unpinned with an inline `hadolint ignore=DL3018` directive explaining why. Other pinned packages still match across both archs and stay pinned.

### Features
- ✨ **i18n coverage extended to the notification outbox, notification rules, and registry/server status badges ([discussion #329](https://github.com/CodesWhat/drydock/discussions/329)).** Four UI surfaces still rendered hardcoded English even under zh-CN: `NotificationOutboxView` (tab labels, table headers, action buttons, toast messages), the notification rule name/description column in `NotificationsView`, and the connection-status badges in `RegistriesView` and `ServersView`. All four are extracted into `t()` catalogs (new `notificationOutboxView` namespace; new entries under `notificationsView.rules.*`, `registriesView.status.*`, and `serversView.status.*`). Rule names/statuses use `te()` so backend-supplied custom names fall back to the raw string when the catalog has no entry. Other locales pick up the new keys on the next Crowdin sync.

## Test plan
- [ ] CI Verify passes
- [ ] After merge: dispatch `release-cut.yml` from main with `release_tag=v1.5.0-rc.21`
- [ ] Confirm rc.21 image lands on GHCR + tag exists + GitHub release published


biggest-littlest (approved):
Approving rc.21 tail (code-owner).

ALARGECOMPANY (approved):
Approving rc.21 tail (code-owner).

PR #376, opened 2026-05-18
🐛 fix(quality): repair mutation testing — UI slices, dashboard publish, Wave 1 mutants

scttbnsn:
## Problem

Run [25990895204](https://github.com/CodesWhat/drydock/actions/runs/25990895204) of the monthly mutation workflow: 18 app slices passed, but both UI slices (`ui-shell-and-data`, `ui-views-and-navigation`) crashed and the dashboard badge has never updated (it returns `"unknown"`).

**Root cause:** the UI slices crash in Stryker's `typescript` checker dry-run compile — `ui/tsconfig.json` has ~35 latent TypeScript errors that normal UI CI never catches (`vite build` + `vitest`/esbuild don't typecheck). With those slices producing no report, the aggregate job fails its `--expected-count 20` check and the `if: success()`-gated dashboard publish is skipped every run. So the dashboard problem is purely downstream of the UI-slice crash.

## Changes

**Infrastructure**
- `fix(ui)` — resolve all 35 UI TypeScript errors, unblocking both UI slices
- `chore(ui)` — add a `tsc --noEmit` typecheck gate to pre-push + CI so latent type errors surface on every push, not once a month via Stryker
- `chore(ci)` — `--allow-missing` on the aggregate step so a single crashed/flaky shard no longer blocks the dashboard publish for a whole month

**Wave 1 mutant kills** — below-threshold slices first
- `openapi` slice: 66.6% → ~94%
- `registries-core` slice: 69.2% → ~88%
- 5 test-only commits; surviving mutants killed across `openapi/common`, `openapi-contract`, paths, `BaseRegistry`, `Registry`, and shared registry helpers; equivalent mutants documented in each commit

## Verification
- App: 8,561 tests pass, 100% coverage
- UI: 3,540 tests pass, 100% coverage; `tsc --noEmit` clean
- Full pre-push gate green (biome, qlty, typecheck-ui, coverage, build, zizmor)

PR #377, opened 2026-05-18
🐛 fix(quality): repair UI mutation slices and dashboard publish visibility

scttbnsn:
Follow-up to #376. With the TypeScript dry-run crash fixed there, the next mutation run ([26010877013](https://github.com/CodesWhat/drydock/actions/runs/26010877013)) exposed two further bugs.

## Bug B — UI slices crash with `No tests were found`

`fix(ui): repair Stryker vitest dry run so UI mutation slices run`

Stryker runs the dry run from a sandbox (`ui/.stryker-tmp/sandbox-XXXXX/`). Two infrastructure test files break there due to relative-path resolution:
- `tests/vitest.coverage-provider.spec.ts` — imports `../app/vitest.coverage-provider.shared.js`, which resolves outside the sandbox.
- `tests/boot/crowdin-config.spec.ts` — resolves the repo root via `../../..`, which lands inside `.stryker-tmp`.

Both throw during the dry run, so Stryker reports `No tests were found` and aborts.

**Fix:** a Stryker-specific `ui/vitest.stryker.config.ts` that extends `vitest.config.ts` and excludes those two infra tests (they cover no `src/` code, so no mutants are hidden); `stryker.conf.mjs` points at it and uses `coverageAnalysis: 'off'` to match the working `app/` config. Verified with a scoped Stryker run — dry run succeeds (3,534 tests) and mutation testing proceeds. `npm run test:unit` and `npm run typecheck` still pass (the regular config is unchanged).

## Bug C — dashboard publish failures reported as success

`fix(ci): surface mutation dashboard publish failures`

The publish retry loop captured `rc=$?` *after* an `if curl ...; fi` compound command — which is always `0` when the condition is false. A failed publish was reported as a successful step. Run 26010877013's publish step "succeeded" while curl actually returned **HTTP 401** on all 3 attempts. Now the step exits 0 only on a real publish, exits 1 after 3 failures, and emits an error annotation.

## ⚠️ Still needed (not in this PR)

The 401 means `STRYKER_DASHBOARD_API_KEY` is invalid/stale and must be regenerated from https://dashboard.stryker-mutator.io. Until then the badge stays `"unknown"` even with these fixes.

## Verification
- Full pre-push gate green (biome, qlty, typecheck-ui, coverage, build, zizmor)
- UI: `typecheck` clean, `test:unit` 3,536 tests pass

### DIFF
--- a/.github/workflows/quality-mutation-monthly.yml
+++ b/.github/workflows/quality-mutation-monthly.yml
@@ -264,6 +264,10 @@ jobs:
           REPO_SLUG: ${{ github.repository }}
           REF_NAME: ${{ github.ref_name }}
         run: |
+          # Each attempt's exit code must come from curl directly. The previous
+          # `rc=$?` ran after the `if` compound command, which is always 0 when
+          # the condition is false — so a failed publish (e.g. HTTP 401) was
+          # silently reported as success. Exit explicitly per outcome instead.
           for attempt in 1 2 3; do
             if curl \
               --fail-with-body \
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
Approved — follow-up fixes verified, full gate green.

ALARGECOMPANY (approved):
Approved — follow-up fixes verified, full gate green.

PR #381, opened 2026-05-18
⚡ perf(ci): speed up mutation testing — perTest + balanced UI slices

scttbnsn:
## Why

The `ui-views-and-navigation` mutation slice projected **~8h** against the 6h
job timeout and inflated the timeout rate. Root cause: PR #377 flipped
`ui/stryker.conf.mjs` `coverageAnalysis` to `'off'`, which re-runs the entire
jsdom suite for *every* mutant instead of only the covering tests.

## What

- **Restore `coverageAnalysis: 'perTest'` in `ui/stryker.conf.mjs`.** `perTest`
  runs only the tests that cover each mutant — Stryker's recommended default.
  Same mutation score, far less wall-clock time. Stryker collects per-test
  coverage via its own instrumentation, independent of the vitest coverage
  provider, so it works inside the mutation sandbox.
- **Align `app/stryker.conf.mjs` to `'perTest'` too** — it was `'off'` with no
  documented reason. Both configs now match what is correct.
- **Split the 2 oversized UI matrix slices into 6 balanced ones**
  (`ui-views-containers`, `ui-views-dashboard`, `ui-composables`,
  `ui-services`, `ui-utils`, `ui-shell`) so each finishes well under the 6h
  job timeout. `--expected-count` bumped 20 → 24.
- **Update `app/mutation-workflow.test.ts`** to expect 24 slices. Its
  coverage/overlap assertion confirms the 6 UI slices cover every UI mutate
  target exactly once with no gaps or overlap.

## Verification

- `perTest` confirmed working in a Linux Stryker sandbox (composables slice
  ran clean).
- `mutation-workflow.test.ts` passes both assertions (count + coverage/overlap).
- Full pre-push gate green: coverage 100%, build, zizmor.


biggest-littlest (approved):
Config-only mutation perf fix; perTest restore + balanced slices verified. LGTM.

ALARGECOMPANY (approved):
Slice split keeps coverage parity (workflow test asserts it). Approved.

PR #389, opened 2026-05-22
🐛 fix(v1.5): rc.26 — image-reference, agent stats, release-notes, and registry-error fixes

### LINKED ISSUE #386: Agents still displaying 0 running containers
### Drydock version

 v1.5.0-rc.24

### What happened?

Basically the same problem, from time to time agents are displayed as having 0 running containers.

Same problem like before (closed ticket): https://github.com/CodesWhat/drydock/issues/362

The agent itself reports 4 detected containers, but the host displays 0. This is not a permanent problem, if I restart it shows correct amount. I haven't detected a pattern of why sometimes it goes to 0.

Image

here are the logs:

Image

I use everywhere latest 
v1.5.0-rc.24

### What did you expect?

Correct display of running containers constantly.

### Steps to reproduce

Just using agents on 
v1.5.0-rc.24

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

_No response_

### OS / Architecture

_No response_

scttbnsn:
Cuts **v1.5.0-rc.26**. Accumulated fixes on `feature/v1.5-rc26` since rc.25.

## Changes

- 🐛 **fix(registries): build image references without corrupting `/v2` path segments**
  `Registry.getImageFullName` and the controller-mode fallback in `resolveContainerImageFullName` stripped `/v2` from the fully concatenated reference with an unanchored regex, silently corrupting an image name that contained a `/v2` segment before it reached Trivy. Extracted a shared `buildImageReference` helper that cleans the registry URL with anchored regexes before concatenation.

- 🐛 **fix(agent): keep the controller agent-summary container count current** ([#386](https://github.com/CodesWhat/drydock/issues/386))
  A cold-start race between the controller handshake and an agent's first watch cycle — plus docker-event-driven changes between crons — left the `AgentsView` running-container count stale. A new `agent-stats-changed` event now fires from every store-mutating path (completed watch cycles, docker-event container changes, the `watch()` path, and the per-container `watchContainer()` path) and is coalesced behind a debounce.

- 🐛 **fix(model): stop flagging a blank "update available" after a transient registry error** ([#342](https://github.com/CodesWhat/drydock/issues/342))
  `hasRawUpdate` compared tags without guarding an undefined `result.tag`, so a mid-flight registry failure (e.g. a 429) flagged the container as updatable with no target version. The tag comparison now runs only when both tags are defined.

- 🐛 **fix(release-notes): survive GitHub's secondary rate limit** ([#342](https://github.com/CodesWhat/drydock/issues/342))
  Release-notes lookups gave up on the first secondary-rate-limit 403. `withRetry` gained opt-in `retryPredicate` / `retryDelayMs` hooks; `GithubProvider` now retries genuine secondary-rate-limit 403s, honours GitHub's retry hint, and arms a bounded module-level cooldown (floored at the 60 s default, capped at 1 h).

- 🐛 **fix(ui): name the failed registry in the Containers registry-error tooltip** ([#342](https://github.com/CodesWhat/drydock/issues/342))
  The registry-error tooltip rendered only the raw transport message; it now derives the registry host from the container's `registryUrl` — e.g. `ghcr.io — Request failed with status code 429`.

- 🔒 **security(deps): override ui `qs@6.15.2` and `js-cookie@3.0.7`** (CVE-2026-8723, CVE-2026-46625)
  Cleared two transitive-dependency CVEs in `ui/package-lock.json` that the qlty osv-scanner gate began flagging. `npm audit` now reports 0 vulnerabilities.

## Verification

- Pre-push gate green on `feature/v1.5-rc26`: biome, qlty (0 security issues), qlty-smells, typecheck, 100% coverage (app 9868 tests + ui 3555 tests, both shards), build, zizmor.
- CHANGELOG stamped `[1.5.0-rc.26] — 2026-05-22`; docs audited — no `content/docs/` changes required.

Closes #386


biggest-littlest (approved):
rc.26 cut — fixes verified, pre-push gate green (qlty 0 issues, 100% coverage app+ui, build, zizmor).

ALARGECOMPANY (approved):
rc.26 cut — approving. CHANGELOG stamped, CVE deps cleared, docs audited.

Issue #385, opened 2026-05-19
Telegram/Pushover: no notification after successful update — `No container found for update-applied` (operation queue / compose recreate)


tarzan77cz:
### Drydock version

 1.5.0-rc.24 (also observed on 1.5.0-rc.20 / rc.23)

### What happened?

After a **successful** container update (auto-update via watcher cron or manual “Update” from the UI), **no lifecycle notification** is sent to Telegram (notification trigger, e.g. `trigger.telegram.*`).

The update itself completes fine (pull, recreate, container running). In logs, the Telegram trigger skips dispatch with:
No container found for update-applied event => ignore

The compose/docker update-action trigger correctly logs (expected since rc.20):
Skipping update-applied dispatch for update-action trigger (lifecycle events do not invoke update lifecycle)

So only notification triggers (Telegram, Pushover, etc.) should handle `update-applied`, but they receive an event **without a `container` in the payload** and the container is **not yet in the store** after compose recreate — same class of race as #355, but for **`update-applied`**, not `update-failed`.

### What did you expect?

A Telegram (or other notification trigger) message when a container update **succeeds** or **fails**, per notification rules for `update-applied` / `update-failed`.

### Steps to reproduce

1. Configure a notification trigger (Telegram) with lifecycle events enabled (`update-applied` / `update-failed` allowed in notification rules).
2. Enable auto-update for a compose-managed container (e.g. LinuxServer image) **or** run manual Update from the UI.
3. Wait for update to complete (image pulled, container recreated and running).
4. Check Telegram — no “update applied” message.
5. Check drydock logs — `No container found for update-applied event => ignore` on the Telegram trigger component.

### Relevant configuration

```yaml
- Update path: `dockercompose.stacks` (compose file under `/app/stacks/...`)
- Auto-updates go through the operation queue (`enqueueContainerUpdate` → `operationId` on trigger)
- Notification trigger: Telegram, `auto: all` (or similar); lifecycle notifications expected regardless of `once` on update-available
```

### Logs

```text
[drydock] 2026-05-17T06:02:31.749 INFO  (drydock/7): Refresh compose service calibre-web from /app/stacks/utils/docker-compose.yml using Docker Engine API {"component":"trigger.dockercompose.stacks","container":"calibre-web"}
[...]

PR #392, opened 2026-05-23
🐛 fix(agent): cross-host update notifications and toasts now fire for agent-side containers (#289)

### LINKED ISSUE #289: Regression with updating container state
### Drydock version

1.5.0-rc.7

### What happened?

When I trigger a manual update of a container it shows queued while updating but that's it. Then after it just disappears from the list of containers with updates. We're missing the "Updating" state. 

### What did you expect?

See above 

### Steps to reproduce

See above 

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

24.0.1

### OS / Architecture

Synology DSM 7

scttbnsn:
Fixes [#289](https://github.com/CodesWhat/drydock/issues/289) — a follow-on cross-host case the rc.20 / rc.25 fixes did not cover.

Reporter [tested rc.25](https://github.com/CodesWhat/drydock/issues/289#issuecomment-4522808507) by running "Update All" on a `tautulli` stack with one instance on his controller host (`datavault`) and one on a connected agent host (`mediavault`). Both updates succeeded, but only the **controller-side** container produced a UI success toast and a Pushover notification; the agent-side container produced neither.

## Cause

When the agent finishes an update it emits `dd:update-applied` over SSE to the controller carrying a full `container` snapshot. The controller's `AgentClient.handleEvent` routes the payload through `maybeMarkAgentOperationSucceededFromAppliedPayload` → `markAgentOperationTerminal` → `ensureAgentOperationForTerminal` → `updateOperationStore.insertOperation` + `markOperationTerminal`. The helper `buildAgentOperationBase` constructed the inserted row from `{id, kind, containerName, containerId, newContainerId}` only — the container snapshot was dropped on the floor.

When `markOperationTerminal` then fired `emitTerminalLifecycleEvent`, `buildTerminalLifecycleEventBase` emitted `emitContainerUpdateApplied({operationId, containerName, containerId})` with **no** container. The notification handler `handleContainerUpdateAppliedEvent` (`app/triggers/providers/Trigger.ts`) then fell back to `findContainerByBusinessId(containerName)`, which compares the agent's bare `containerName` (`tautulli`) against the controller-side `fullName` (`mediavault_docker_tautulli`) — and silently dropped. Same shape on the `dd:update-failed` branch via `maybeMarkAgentOperationFailedFromFailedPayload`.

This is the same class of `findContainerByBusinessId` miss as #385 but on the agent-scoped operation path that #385 did not cover.

## Fix

Thread the agent's container snapshot through every level of the agent-scoped operation pipeline so the store's terminal-lifecycle emit naturally carries it:

- `buildAgentOperationBase` accepts an optional `container` and stamps `agent: this.name` on it (caller wins — the controller's view always overrides any inbound `agent` field).
- `ensureAgentOperationForTerminal` patches the container onto an **existing active** row (the `dd:update-operation-changed`-before-`dd:update-applied` race) via `updateOperation`, but only when the existing row lacks a snapshot — never overwrites.
- `markAgentOperationTerminal` / `maybeMarkAgentOperationSucceededFromAppliedPayload` / `maybeMarkAgentOperationFailedFromFailedPayload` thread the container through.
- `container` is added to `MutableUpdateOperationFields` in `app/store/update-operation.ts` so terminal and active patches accept it.

No changes to the agent SSE serializer or to the controller-side dispatch path — the fix is upstream of the emit so the existing `payloadContainer` shortcut in `handleContainerUpdateAppliedEvent` now succeeds for agent-originated updates. The early `return` in `AgentClient.handleEvent` is preserved, so there is no double-emit risk.

## Tests

`app/agent/AgentClient.test.ts`:
- Updated the existing `should terminalize agent update-applied payloads with operation ids through the store` test to assert `insertOperation` and `markOperationTerminal` are now called with `container.agent === 'test-agent'`.
- Added: regression test for #289 (agent applied with operationId AND container threads container through both insert and terminal patch).
- Added: agent-name stamp override regardless of inbound `container.agent`.
- Added: existing-row race — `dd:update-operation-changed` already inserted an active row without container, then `dd:update-applied` arrives; the active row is patched via `updateOperation` with the container snapshot before the terminal emit runs.
- Added: `dd:update-failed` with container threads container into `markOperationTerminal`.

## Verification

- Pre-push gate green: biome, qlty (0 security issues), qlty-smells, typecheck-ui, 100% coverage (9872 app tests + ui shard), build.
- CHANGELOG entry under `[Unreleased]` so the rc.27 release notes pick this up alongside the rc.26 batch.

Closes #289


biggest-littlest (approved):
Approve: Diagnosis matches the symptom in the reporter's recording — only one toast appears for two updated containers because the agent-scoped operation row carries no container snapshot into the terminal lifecycle emit, dropping the trigger handler's lookup. Threading container through buildAgentOperationBase / ensureAgentOperationForTerminal is the right shape; the existing-row `updateOperation` patch closes the dd:update-operation-changed-first race without overwriting. Tests cover both branches and the race.

ALARGECOMPANY (approved):
Approve: `container` addition to `MutableUpdateOperationFields` and `Pick`-extended patch types are safe — no existing call site is constrained on the exact field set, and the `persistOperationPatch` spread accepts the new field at runtime. `agent: this.name` stamp prevents an inbound spoofed agent name from confusing controller-side container resolution. CHANGELOG entry is under Unreleased so rc.27 picks it up.

scttbnsn:
Superseded by #393.

The two commits on this branch (`a7f633b7` — container snapshot threading, `2132183e` — CHANGELOG entry) are both present on `feature/v1.5-rc27` and ship with #393, alongside the follow-on `operationId` end-to-end threading fix and the CodeQL-clearing tests. Closing in favor of the consolidated rc.27 PR.

PR #393, opened 2026-05-23
🐛 fix(v1.5): rc.27 — agent-hosted update operationId threading + container snapshot (#289)

### LINKED ISSUE #289: Regression with updating container state

scttbnsn:
Closes the long-running same-name / cross-host update-state regression on [#289](https://github.com/CodesWhat/drydock/issues/289). Two coordinated fixes on top of rc.26:

## 1. Thread the agent's container snapshot through every level of the agent-scoped operation pipeline (`a7f633b7`)

Cross-host updates silently dropped notification triggers (Pushover, Telegram, etc.) and produced incomplete SSE toasts for containers running on a connected agent. The controller built the agent-scoped operation row from `buildAgentOperationBase` without the container snapshot, so when `markOperationTerminal` fired `emitTerminalLifecycleEvent`, the resulting `emitContainerUpdateApplied` / `emitContainerUpdateFailed` carried only `{operationId, containerName, containerId}`. The notification trigger handler then fell back to `findContainerByBusinessId`, which compares the agent's bare `containerName` (`tautulli`) against the controller's `fullName` (`mediavault_docker_tautulli`) and silently dropped — the same class of `findContainerByBusinessId` miss as #385 but on the agent-scoped operation path.

Threads `container` through `buildAgentOperationBase` / `ensureAgentOperationForTerminal` / `markAgentOperationTerminal` / `maybeMarkAgentOperationSucceededFromAppliedPayload` / `maybeMarkAgentOperationFailedFromFailedPayload`, stamping `agent: this.name`. Existing-row race (`dd:update-operation-changed` before `dd:update-applied`) patches `container` onto the active row via `updateOperation` before the terminal emit. Adds `container` to `MutableUpdateOperationFields` so patch types accept it.

## 2. Thread the controller's `operationId` end-to-end so agent-hosted updates use exactly one row (`d38ea2c7`)

Even with fix #1, a follow-up rc.25 video showed a `[mediavault] Container Tautulli update failed — Marked failed after exceeding active update TTL (1800000ms) while queued.` Pushover arriving ~30 minutes after a successful update. Cause: `createAcceptedContainerUpdateRequest` mints a controller-side queued row with a `crypto.randomUUID()` `operationId`, but `AgentTrigger.trigger(container)` discarded the `runtimeContext` so the operationId never reached the agent. The agent minted its own UUID and fired lifecycle events keyed by that id, which the controller routed through `toAgentScopedId` into a third (agent-scoped) row. The original controller-side queued row was therefore never touched, sat past `UPDATE_OPERATION_ACTIVE_TTL_MS`, and was force-failed by the TTL sweep — which fired the misleading "failed" notification with the row's still-valid container snapshot.

Threads the controller `operationId` end-to-end:

- `AgentTrigger.trigger` / `triggerBatch` accept and forward `runtimeContext`
- `AgentClient.runRemoteTrigger` / `runRemoteTriggerBatch` extract per-container operationIds via `getRequestedOperationId` and include them in the agent payload (`{id, name, operationId}` single; `{...container, operationId}` per entry for batches)
- `app/api/trigger.ts runTrigger` accepts an `operationId` in the body (validated by `triggerRequestBodySchema`) and threads it into `requestContainerUpdate`
- `app/agent/api/trigger.ts runTriggerBatch` extracts per-container operationIds into `{operationIds}` runtimeContext before forwarding to the local trigger
- `EnqueueContainerUpdateOptions` gains an `operationId` field honored by `createAcceptedContainerUpdateRequest` (single-container batches only; multi-container batches still mint per-container UUIDs)
- New `AgentClient.resolveAgentOperationId` checks the controller's operation store for an existing row at the raw (unscoped) id and reuses it when found, falling back to `toAgentScopedId` when the agent does not echo a known controller id (backwards-compat with older agents)

The controller-side queued row therefore transitions directly to `in-progress` and `succeeded`/`failed` from the agent's lifecycle events, no parallel agent-scoped row is created, the TTL sweep has nothing stale to fail, and the spurious "update failed" notification disappears.

## Tests / coverage

- 9903 app tests pass with 100% coverage (statements / branches / functions / lines)
- New tests cover: `AgentTrigger` runtimeContext forwarding, `runRemoteTrigger`/`Batch` operationId payload, agent `runTrigger`/`runTriggerBatch` operationId extraction, `request-update` operationId honoring (single-container only), and `AgentClient.resolveAgentOperationId` reuse-vs-scope behavior
- Build clean, lint clean (no new biome findings)

Fixes: #289


biggest-littlest (approved):
Approving — closes the two #289 regressions: (1) agent container snapshot threading so notification triggers resolve agent-hosted containers; (2) controller operationId threaded end-to-end so the orphaned queued row + spurious TTL-failure notification is eliminated. 100% coverage held.

biggest-littlest (dismissed):
Approving — closes the two #289 regressions: (1) agent container snapshot threading so notification triggers resolve agent-hosted containers; (2) controller operationId threaded end-to-end so the orphaned queued row + spurious TTL-failure notification is eliminated. 100% coverage held.

ALARGECOMPANY (approved):
Approving — operationId threading is clean (single source of truth for the lifecycle row), and the backwards-compat fallback in resolveAgentOperationId preserves older-agent behavior. Coverage held.

ALARGECOMPANY (dismissed):
Approving — operationId threading is clean (single source of truth for the lifecycle row), and the backwards-compat fallback in resolveAgentOperationId preserves older-agent behavior. Coverage held.

biggest-littlest (approved):
Re-approving after CodeQL fix (af82937e) — null-prototype guard + forbidden-key validation on the new operationIds map closes the js/remote-property-injection finding. Added tests cover the forbidden-key and non-string-id branches. Coverage held.

biggest-littlest (dismissed):
Re-approving after CodeQL fix (af82937e) — null-prototype guard + forbidden-key validation on the new operationIds map closes the js/remote-property-injection finding. Added tests cover the forbidden-key and non-string-id branches. Coverage held.

ALARGECOMPANY (approved):
Re-approving after CodeQL fix — null-prototype + forbidden-key guard is the right pattern. Tests cover the new branches.

ALARGECOMPANY (dismissed):
Re-approving after CodeQL fix — null-prototype + forbidden-key guard is the right pattern. Tests cover the new branches.

biggest-littlest (approved):
Re-approving after CodeQL Map fix (b76e06d9). Per CodeQL's own QHelp for js/remote-property-injection, Map is the canonical mitigation — Map writes aren't recognized as a sink. Consumer accepts Map | Record with Object.hasOwn + Reflect.get on the Record path as defense-in-depth. Coverage 100%.

biggest-littlest (dismissed):
Re-approving after CodeQL Map fix (b76e06d9). Per CodeQL's own QHelp for js/remote-property-injection, Map is the canonical mitigation — Map writes aren't recognized as a sink. Consumer accepts Map | Record with Object.hasOwn + Reflect.get on the Record path as defense-in-depth. Coverage 100%.

ALARGECOMPANY (dismissed):
Re-approving — Map is the right call per CodeQL docs. Defense-in-depth on the legacy Record path is a nice bonus.

biggest-littlest (approved):
Re-approving after rc.27 test-fix + README polish. CodeQL js/invalid-prototype-value cleared by 6a9b9e69 (replace the `__proto__: 'evil-proto'` object-literal — which is silently dropped by the engine — with `Object.create(prototypeWithDataProp)` so the consumer's Object.hasOwn guard is genuinely exercised against an inherited data property). 6c4fce63 is docs-only README cleanup. Coverage held at 100%.

ALARGECOMPANY (approved):
Re-approving — fix is the right shape (Object.create(prototypeWithDataProp) actually exercises the hasOwn guard against an inherited data property; the previous `__proto__: 'evil-proto'` literal was silently dropped). Docs-only README polish on top. 100% coverage held.

PR #396, opened 2026-05-27
🚀 release(v1.5): rc.28 — handshake cold-start race + perf/security/refactor batch

scttbnsn:
## Summary

Cuts v1.5.0-rc.28 — 19 commits since rc.27, all on `feature/v1.5-rc28`.

### Highlights
- **#386 follow-through (`7ef4aa61`)** — `AgentClient._doHandshake` now defers `pruneOldContainers` when handshake returns 0 containers, eliminating the cold-start race where a freshly-restarted agent would wipe the controller's last-known container count between handshake and first watch cycle. Combined with the rc.25 (`d02080ae`) snapshot suppression and rc.26 (`512c3751`) stats-changed broadcast, the agent count display is now robust across the full restart lifecycle. Cesc1986's rc.27 reporter log proved the failure mode (`agent-client.ml Handshake → 0` after restart, wiping 5 known containers).
- **#386 concurrent-handshake guard (`3da42587`)** — in-flight `_doHandshake` Promise reused across overlapping reconnects.
- **Security caps (`5e02f6df`, `9b6e1b97`)** — auth JSON body limit + global SSE connection cap; doc note that Command trigger inherits all `DD_*` env.
- **Perf batch (`07e3580e`, `b2999eb7`, `dceec898`)** — store/watcher/SSE replay hot paths, bundle splitting, SSE jitter, OpenAPI cache, alert fan-out, containers menu scan, WS log backpressure.
- **Refactors (`47c0cf44`, `4d34b5e7`, `6ee07006`)** — notifications outbox openapi docs + drain-safe stop, registry contract hygiene, `readJsonResponse` adoption.
- **UI fix (`de07d412`)** — include scanning/sbom-generating in phase enum.
- **CI polish (`794497e5`, `06e56be3`, `612e81c4`, `06eed8dc`, `82caeb08`, `3721f66b`)** — stub `getContainersRaw` in same-source watcher tests, vite codeSplitting i18n group, `runTrigger` json type narrowing, Crowdin Sync emoji prefix, transient 401 checkout retry, exclude pure-data files from UI mutation slice.

### CHANGELOG
- `[Unreleased]` now contains only the #386 deferred-prune entry (will be stamped to `[1.5.0-rc.28]` by `release-cut.yml`).
- `[1.5.0-rc.27]` section added retroactively for the two #289 entries that shipped in rc.27 release notes but were never stamped into the file.

## Test plan
[...]

PR #397, opened 2026-05-28
🔧 chore(ci): cache Playwright browsers + bump npm-ci timeout to 15m

scttbnsn:
## Summary

Three consecutive Playwright runs failed at the same step on `cdn.playwright.dev` CDN throttling:
- rc.28 PR #396 first attempt (`bca0b7bb`)
- rc.28 PR #396 second attempt (`bca0b7bb`)
- post-merge on main (`7dc8b2d3`) — blocking release-cut for v1.5.0-rc.28
- post-merge retry attempt 2 — same shape

Each attempt: chromium download reaches 100% at ~9m30s, then `node install.js` SIGTERMs at the 10-min wall clock. Same workflow file, same package versions — CDN throttling is currently routing each runner pool through a slow edge.

### Changes

- **actions/cache for `~/.cache/ms-playwright`** keyed on lockfile hash + `github.ref_name`. Each branch gets its own cache namespace (matches the scoping pattern in `quality-mutation-monthly.yml`). After the first cold-cache run, subsequent runs short-circuit the postinstall download.
- **timeout_minutes 10 → 15** covers the worst observed throttled-download case with margin for the `npm install` post-extraction work.
- Inline `# zizmor: ignore[cache-poisoning]` pragma with rationale; the workflow's push:main + PR triggers cause zizmor to flag any actions/cache as a poisoning risk regardless of key scope.

The dedicated `Install Playwright Chromium` step that runs after `npm ci` keeps its 10-min retry budget and will now hit the warm cache on every run.

## Test plan
- [x] Pre-push gates green locally (clean-tree, biome, qlty, qlty-smells, typecheck, 100% coverage, build, zizmor)
- [x] zizmor passes on the edited workflow
- [ ] PR-event Playwright workflow run passes on this commit
- [ ] After merge: rerun the failed main-branch Playwright on `7dc8b2d3` (which won't have the fix on its tree but will share the cache — chromium binary persists), OR re-dispatch release-cut against the new merge commit

### DIFF
--- a/.github/workflows/e2e-playwright.yml
+++ b/.github/workflows/e2e-playwright.yml
@@ -130,10 +130,38 @@ jobs:
           cache-from: type=gha,scope=drydock
           cache-to: type=gha,mode=max,scope=drydock,ignore-error=true
 
+      # Cache Playwright's browser binary directory across runs.
+      # The @playwright/browser-chromium devDep runs a 167-MiB postinstall
+      # download against cdn.playwright.dev during `npm ci`. CDN throttling
+      # has been making that 9-10 min per attempt and three jobs in a row
+      # blew past even a 10-min retry timeout. With a warm cache the
+      # postinstall short-circuits and npm ci finishes in seconds.
+      # Per-ref key matches the cache-scoping pattern in
+      # quality-mutation-monthly.yml: each branch (main, release/**, PR
+      # heads) gets its own cache namespace so an untrusted PR cannot
+      # poison main's chromium binary. restore-keys lets a fresh branch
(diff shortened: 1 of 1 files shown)


biggest-littlest (dismissed):
Approving: cache + timeout fix for the chromium-download CDN throttle.

ALARGECOMPANY (dismissed):
LGTM.

biggest-littlest (approved):
Re-approving after zizmor pragma placement fix.

ALARGECOMPANY (approved):
LGTM, pragma now adjacent to the flagged line.

PR #401, opened 2026-05-28
🔧 chore(ci): mark DAST advisory while we debug post-rc.28 401 race

scttbnsn:
## Summary
- DAST job (ZAP + Nuclei) reliably 401s on \`POST /auth/login\` against the post-rc.28 QA stack despite /auth/strategies returning a populated array.
- Cucumber e2e doesn't trip it because its first scenario polls /auth/strategies, letting init settle by the time it hits /auth/login.
- DAST is currently blocking release-cut for rc.28. Mark advisory so the CI Verify conclusion is success while we debug the auth init path separately. Findings still upload as SARIF + artifacts.

## Test plan
- [ ] CI Verify on main goes green
- [ ] Then: \`gh workflow run release-cut.yml --ref main -f release_tag=v1.5.0-rc.28\`
- [ ] Open follow-up to root-cause the auth init 401 race

### DIFF
--- a/.github/workflows/ci-verify.yml
+++ b/.github/workflows/ci-verify.yml
@@ -523,7 +523,7 @@ jobs:
           retention-days: 1
 
   dast-zap-baseline:
-    name: "🕷️ DAST: ZAP + Nuclei"
+    name: "🕷️ DAST: ZAP + Nuclei (Advisory)"
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


biggest-littlest (approved):
✅ Approve advisory-DAST while we debug.

ALARGECOMPANY (approved):
✅ Approve.

PR #402, opened 2026-05-29
🐛 fix(auth): close strategy-registration gap so login is ready before /health is

scttbnsn:
## Summary

Fixes the post-rc.28 DAST 401 race where `POST /auth/login` would fail immediately after `/health` returned 200, because `STRATEGY_IDS` in `auth-strategies.ts` was still empty when the request arrived.

**Root cause:** `passport.authenticate([], cb)` with an empty array calls `allFailed()` immediately → `callback(null, false, [], [])` → 401. The `/health` endpoint (previously `express-healthcheck`) returned 200 as soon as Express was bound, regardless of whether `auth.init(app)` → `registerStrategies()` had run.

**Fix:** Gate `/health` on a live `getAllIds().length > 0` check. The endpoint returns `503 {"status":"starting","reason":"auth strategies not yet registered"}` until passport strategies are registered, then `200 {"uptime":...}`. DAST (and any readiness poller) will now block at `/health` until the auth chain is fully wired.

- `app/api/health.ts` — replaced `express-healthcheck` with a custom `healthHandler`; added `setAuthReadyFn()` and `resetAuthReadyFnForTests()` exports
- `app/api/index.ts` — wires `healthRouter.setAuthReadyFn(() => getAllIds().length > 0)` before `auth.init(app)` in `registerRoutes()`
- `app/api/health.test.ts` — full coverage of 200/503 behaviour and both exported helpers
- `app/api/index.test.ts` — asserts `setAuthReadyFn` is called before `auth.init` and that the closure is a live `getAllIds()` check
- `.github/workflows/ci-verify.yml` — removes the `continue-on-error` advisory flag and the `/auth/strategies` polling workaround from "Wait for QA health"; restores the DAST job name

## Test plan

- [ ] `app/` vitest: 355 files, 9938 tests — all pass (verified locally)
- [ ] UI vitest: passes
- [ ] `.github/tests` workflow tests: 8 files, 26 tests — all pass
- [ ] Pre-push hooks: clean-tree, biome, qlty, coverage (100%), build, zizmor — all green
- [ ] CI: DAST job should pass without `continue-on-error`; `/health` returns 503 until strategies are registered, then 200 — login succeeds

### DIFF
--- a/.github/workflows/ci-verify.yml
+++ b/.github/workflows/ci-verify.yml
@@ -523,7 +523,7 @@ jobs:
           retention-days: 1
 
   dast-zap-baseline:
-    name: "🕷️ DAST: ZAP + Nuclei (Advisory)"
+    name: "🕷️ DAST: ZAP + Nuclei"
     runs-on: ubuntu-latest
     timeout-minutes: 25  # Includes QA stack startup and scanner container runtime
     # Run on every merge to main (gates the next release-cut), on legacy
@@ -533,9 +533,6 @@ jobs:
       (github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')))
       || github.event_name == 'schedule'
     needs: [build]
-    # Advisory while we debug a post-rc.28 401 race in the auth init path —
-    # findings still surface in SARIF + artifacts; release-cut isn't gated.
-    continue-on-error: true
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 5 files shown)


biggest-littlest (approved):
LGTM — structural fix matches the diagnosed race in registerAuthentications. Playwright in-vivo pass confirms the readiness gate works.

ALARGECOMPANY (approved):
Approved. CodeQL false-positive on the /health rate-limit rule was dismissed with reasoning (liveness probes shouldn't be rate-limited).

PR #404, opened 2026-05-29
🐛 fix(ci): relativise ZAP SARIF URIs with originalUriBaseIds for code scanning

scttbnsn:
## Summary

After PR #402 (auth-readiness gate) and PR #403 (correct credentials), DAST now scans successfully — but Upload ZAP SARIF still fails:

> Code Scanning could not process the submitted SARIF file: SARIF URI scheme \"http\" did not match the checkout URI scheme \"file\"

Per SARIF spec §3.14.14, DAST results must store origins in \`originalUriBaseIds\` and use relative URIs with \`uriBaseId\` references. The current converter emits raw \`http://localhost:3333/...\` URIs in \`artifactLocation.uri\`, which Code Scanning rejects.

## Changes

\`scripts/zap-json-to-sarif.mjs\`:
- New \`resolveArtifactLocation()\` helper detects http(s) URIs, strips the origin, returns \`{ uri: relativePath, uriBaseId: 'TARGET' }\`. Non-http fallback values (\`nodeName\`, site \`@name\`, \`'zap-target'\`) pass through unchanged with no \`uriBaseId\`.
- \`convertZapJsonToSarif()\` collects distinct origins into \`originalUriBaseIds\` at the run level.

\`scripts/zap-json-to-sarif.test.mjs\` updated assertions for the new shape.

## Verification

- \`node --test scripts/zap-json-to-sarif.test.mjs\` → 10/10 pass
- Pre-push gates all green (coverage, build, biome, qlty)
- Post-merge CI Verify run 26613337624 confirms this is the last blocker: every DAST step passed except \`Upload ZAP SARIF\`

## Test plan
- [ ] CI Verify on this PR passes lint/coverage/scripts-test
- [ ] Post-merge CI Verify on main: DAST → Upload ZAP SARIF succeeds (the canonical proof)

PR #405, opened 2026-05-29
🐛 fix(ci): preserve leading slash in ZAP SARIF artifact URIs so root scans validate

scttbnsn:
## Summary

Follow-up to PR #404. The previous fix relativised http URIs but stripped the leading \`/\`, which meant root-path scans (\`http://localhost:3333/\` — the most common ZAP baseline target) ended up as \`uri: ''\`. GHAS Code Scanning then rejected the upload:

> \`locationFromSarifResult: expected artifact location\`

(seen in post-merge CI Verify run 26615030359, job 78430341466)

## Fix

Don't strip the leading slash. Result: root URL emits \`uri='/'\`, paths emit \`uri='/auth/login'\` etc. Both are valid absolute-path-references per RFC 3986 §4.2 and resolve correctly against \`originalUriBaseIds.TARGET\` per SARIF §3.4.4.

## Verification
- 10/10 ZAP tests pass; 53/53 full scripts suite pass
- Pre-push gates all green (coverage 100%, build, biome, qlty)

## Test plan
- [ ] Post-merge CI Verify on main: \`Upload ZAP SARIF\` succeeds — fully closes the DAST chain

### DIFF
--- a/scripts/zap-json-to-sarif.mjs
+++ b/scripts/zap-json-to-sarif.mjs
@@ -154,9 +154,11 @@ function resolveArtifactLocation(rawUri) {
   try {
     const parsed = new URL(rawUri);
     if (parsed.protocol === 'http:' || parsed.protocol === 'https:') {
-      // Relative path (without leading slash) + uriBaseId referencing the origin.
-      const relative = (parsed.pathname + parsed.search + parsed.hash).replace(/^\//, '');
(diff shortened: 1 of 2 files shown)


biggest-littlest (approved):
LGTM — preserves leading slash, satisfies absolute-path-reference per RFC 3986.

ALARGECOMPANY (approved):
Approved.

PR #417, opened 2026-06-06
🔧 chore(ci): pin markdownlint + bounded test retry for local/CI determinism

scttbnsn:
Two security-neutral reliability fixes that came out of the rc.33 cut. Neither touches branch protection or security scanning.

## 1. Pin markdownlint to 0.46.0 (`.qlty/qlty.toml`)

markdownlint was the only lint plugin left unpinned (shellcheck and trivy are already pinned). Unpinned, qlty resolves whatever markdownlint the installed CLI defaults to, which drifts between a developer’s local qlty and CI’s latest qlty. That drift let a docs change pass `qlty check --all` locally and then **fail the CI lint gate on MD032** during the rc.33 cut. Pinning makes the markdownlint ruleset identical in both places.

Validated locally: with the pin, `qlty check` flags an MD032 violation and passes the real doc clean.

## 2. Bounded vitest retry (`app/vitest.config.ts`, `retry: 1`)

Guards against a rare, non-deterministic cross-file isolation flake: under serial execution a prior file’s module mock occasionally leaks into `registry/index.test.ts` (passes 98/98 in isolation; not reproducible on demand; CI passes on the same SHA). `retry: 1` only re-runs a *failing* test once, so green runs and consistently-broken tests are unaffected — it cannot mask a real failure. The underlying leak is tracked separately.

## Not included (deliberately)

Relaxing branch protection and moving CodeQL/scans off per-PR runs were considered and **dropped** — both would lower the OpenSSF/security posture.

### DIFF
--- a/.qlty/qlty.toml
+++ b/.qlty/qlty.toml
@@ -80,6 +80,11 @@ name = "hadolint"
 [[plugin]]
 name = "markdownlint"
 mode = "comment"
+# Pinned so the markdownlint ruleset is identical locally and in CI. Left
+# unpinned, qlty resolves whatever markdownlint the installed CLI defaults to,
+# which drifts between a developer's qlty and CI's (latest) qlty — a doc that
+# passed `qlty check --all` locally then failed the CI lint gate on MD032.
+version = "0.46.0"
 
 [[plugin]]
 name = "osv-scanner"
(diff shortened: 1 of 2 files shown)


biggest-littlest (approved):
LGTM — pins the floating markdownlint plugin to kill local/CI drift; retry:1 is a safe, bounded guard for the known isolation flake. No security posture change.

ALARGECOMPANY (approved):
Approved. Determinism-only: plugin version pin + single test retry. Branch protection and security scans untouched.

Issue #421, opened 2026-06-09
False "update failed" notification due to 409 Conflict race condition before successful update


tarzan77cz:
### Drydock version

1.5.0-rc34

### What happened?

We are receiving a false "update failed" notification followed immediately by a successful update notification when a remote agent updates a container.

This seems to be a race condition related to the fixes introduced in #410 and #411. While the classifyDuplicateOpTerminalStatus logic attempts to silence 409 Conflict errors by marking them as expired rather than failed, it relies on getRecentTerminalSucceededOperationByContainerName finding a recently succeeded operation.

If the agent returns the HTTP 409 response before the controller has fully received and processed the dd:update-applied event (or while the update is still actively in-progress on the agent), the conflict is classified as failed instead of expired, resulting in a false alert.

### What did you expect?

The 409 Conflict (which originates as an Axios error: "Request failed with status code 409") should be correctly classified as a benign duplicate (expired) even if the active update is still in-progress and hasn't yet produced a terminal succeeded state in the controller's store.

In duplicate-op-classification.ts, the classifyDuplicateOpTerminalStatus function could be enhanced to also return 'expired' if there is currently another active (in-progress) operation for the exact same container name and identity, or if the 409 error explicitly indicates an active update lock, rather than relying exclusively on a completed success state.

### Steps to reproduce

1. Set up a Drydock controller with at least one connected remote agent (e.g., schladming).
2. Have a container running on the remote agent (e.g., zigbee2mqtt_schladming) with an available image update.
3. Trigger an update for this container multiple times concurrently (this can happen naturally due to a race condition between the controller's watch loop and a webhook, or manually by rapidly clicking the "Update" button twice in the UI).
4. The controller sends multiple POST requests to the agent's /api/triggers/... endpoint.
5. The agent accepts the first request and starts the update process (creating an active operation).
6. The agent correctly rejects the subsequent concurrent requests with an HTTP 409 Conflict ("Container update already in progress").
7. The controller's Axios client throws Request failed with status code 409 for the rejected requests.
8. The controller catches this Axios error, marks the duplicate operation as failed, and incorrectly fires an update-failed notification.
9. Shortly after, the original update on the agent finishes successfully, and the controller fires an update-applied (updated successfully) notification.

### Relevant configuration

```yaml

```

### Logs

```text
[10.06.2026 0:00] Tarzan Update Bot: [schladming] New tag found for container zigbee2mqtt_schladming (stacks)
[...]

Issue #410, opened 2026-06-03
False "update failed (404 / 409)" notifications after successful auto-update


tarzan77cz:
### Drydock version

1.5.0-rc.29

### What happened?

When using automatic updates via dd.action.include (e.g., dockercompose.stacks), the update successfully completes and sends an updated successfully notification. However, a few minutes later, ghost update failed notifications arrive with HTTP 404 (No such container) or HTTP 409 (Conflict).

The containers are updated correctly and running fine, but the notification system spams failures for dead operations.

### What did you expect?

If an auto-update succeeds, no secondary "ghost" operations should fail and trigger update-failed notifications.

### Steps to reproduce

It looks like a race condition where a container report is enqueued multiple times (e.g. from both handleContainerReport and handleContainerReports, or from subsequent watcher cron passes before the first operation clears the queue).

The first operation succeeds and recreates the container (new ID).
The duplicate operation eventually executes but holds the snapshot with the old Docker ID.
It hits the Docker API (gets 404) or agent API (gets 409), the operation is marked terminal as failed, and Trigger.ts blindly emits the update-failed notification to Telegram/Pushover.

### Relevant configuration

```yaml

```

### Logs

```text
Logs / Symptoms:

[03.06.2026 12:00] New tag found for container n8n
[03.06.2026 12:05] Container prowlarr updated successfully
[03.06.2026 12:06] Container n8n updated successfully
[03.06.2026 12:11] Container n8n update failed: (HTTP code 404) no such container - No such container: 33f037...
(Similarly, on agent-hosted containers, it throws Request failed with status code 409 instead of 404).
```

### Installation method

Docker Compose

### Docker version

29.3.0

### OS / Architecture

Linux docker 6.11.11-2-pve

scttbnsn:
Hi @tarzan77cz — this is fixed in **[v1.5.0-rc.30](https://github.com/CodesWhat/drydock/releases/tag/v1.5.0-rc.30)**, just published.

**Root cause:** after a successful update, a *duplicate* update operation could run against a container that had already been recreated (by a previous pass, Docker Compose, or an agent), hitting a genuine Docker `404 no such container` / `409 conflict`. That was terminalised as `failed` and fired a false "update failed" notification — even though your update had actually succeeded.

[...]

PR #424, opened 2026-06-11
🐛 fix(store): normalise empty-string identity fields in insertOperation

scttbnsn:
## Summary

Hardening batch from today's v1.5.0 release-readiness review of the rc.33→rc.35 delta. No behavior changes for any current caller — this makes the duplicate-409 race fix's identity-matching invariant unconditional and closes the three test gaps the review flagged around #421/#422.

### Fix

- **`insertOperation` now normalises `''` → `undefined`** for `agent` and `watcher`, both top-level and inside the `container` snapshot. Previously the comment in `matchesStrictIdentityFilter` claimed this normalization existed, but it didn't — the invariant only held because every current caller filters empty strings before insert. A future caller passing `agent: ''` would have silently broken strict identity matching (the cross-agent guard in the 409-race classification). The container snapshot is **cloned on normalize**, never mutated in place, since it belongs to the caller and may be a live store document.

### Tests added

- **TTL-expired winner** (`store/update-operation.test.ts`): a racing "winner" op stored as `in-progress` but past `DD_UPDATE_OPERATION_ACTIVE_TTL_MS` — `hasOtherActiveOperationByContainerName` returns `false` and the freshness check terminalizes the stale op as `expired`.
- **Ambiguous same-name containers** (`agent/AgentClient.test.ts`): two agent-owned containers sharing a name — `getStoredContainerForAgentOperation` returns `undefined` and `buildAgentOperationBase` stamps no container/watcher identity, instead of picking one arbitrarily.
- **Winner fails after loser expired** (`updates/request-update.test.ts`): sequences loser-409→`expired` then winner→`failed`; asserts the final store state is exactly one `failed` + one `expired` and the update-failed terminal transition happens exactly once — the exact observable behavior #410's reporter will judge rc.35 by.
- **Normalization coverage** (`store/update-operation.test.ts`): empty-string/mixed/preserved identity cases, plus an assertion that the caller's container object is not mutated.

## Verification

- Full pre-push gate passed: 100% coverage thresholds met (app + ui), tsc clean, biome clean, qlty clean, builds clean.
- 663 tests passing across the three affected suites.

### DIFF
--- a/app/agent/AgentClient.test.ts
+++ b/app/agent/AgentClient.test.ts
@@ -4841,6 +4841,54 @@ describe('AgentClient', () => {
       expect(result).toBe(row);
       expect(storeContainer.getContainers).toHaveBeenCalledWith({ agent: 'test-agent' });
     });
+
+    test('returns undefined when two agent-owned containers share the same name (ambiguous match)', () => {
+      // Two containers both named 'nginx' on this agent — cannot determine which
+      // identity to stamp, so the function must return undefined rather than
+      // picking one arbitrarily.
+      vi.mocked(storeContainer.getContainer).mockReturnValue(undefined);
+      vi.mocked(storeContainer.getContainers).mockReturnValue([
+        { id: 'nginx-1', name: 'nginx', agent: 'test-agent', watcher: 'local' },
+        { id: 'nginx-2', name: 'nginx', agent: 'test-agent', watcher: 'local' },
+      ] as any);
+
+      const result = (client as any).getStoredContainerForAgentOperation({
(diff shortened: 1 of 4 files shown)


biggest-littlest (approved):
Reviewed the normalization change and the three new race-coverage tests — clone-on-normalize is the right call for the caller-owned snapshot. LGTM.

ALARGECOMPANY (approved):
Identity invariant now enforced at the store boundary instead of by caller discipline, and the winner-fails-after-loser-expired sequence finally has a regression test. LGTM.

PR #425, opened 2026-06-11
🐛 fix(self-update): hold UI overlay until the container swap is complete

scttbnsn:
## Problem

During a self-update the "Applying Update" overlay flickers: it appears, the app comes back almost immediately, drops again when the old container actually stops, comes back showing the transient `drydock-self-update-<timestamp>` helper container in the list, and only then settles. Three compounding defects:

1. **UI dismissed the overlay on any server response.** The connectivity probe polled `/auth/user` and treated any 200/401 as "update finished" — but the old server keeps answering throughout the image pull. Separately, the browser's **native EventSource retry** (never closed in self-update mode) reconnected to the new server the moment it listened, fired `sse:connected`, and cleared the overlay early — leaving the SPA on stale pre-update assets.
2. **Completion was never recorded.** The finalize-callback secret was per-process, so the helper's finalize POST always 403'd after the restart, and startup reconciliation expired the in-progress operation before the helper could finalize it.
3. **The helper container had no watch-exclusion label**, so watch-by-default pulled it into the container list until it auto-removed.

## Fix

- New unauthenticated `GET /api/v1/self-update/:operationId/status` endpoint (operation UUID as the access capability, minimal response, global rate limit) — the UI polls it during a self-update and hard-reloads only when the operation reaches a terminal status (`succeeded` / `rolled-back` / `failed` / `expired`).
- Per-operation finalize secret with its SHA-256 hash persisted on the operation row, so the restarted process validates the helper's callback (legacy rows fall back to the process secret).
- Startup reconciliation grants fresh in-progress self-update operations a 10-minute grace window; the status endpoint expires overdue ones, and the grace clock restarts when the helper spawns so a slow pull can't eat the finalize budget.
- The UI closes its SSE connection when self-update mode begins (after the ack is delivered) — the status-poll-then-reload path is the only recovery during a self-update.
- Helper container is labeled `dd.watch=false`.
- Dry-run no longer shows the overlay and marks the no-op operation terminal.
- Docs (self-update page, API table, OpenAPI), CHANGELOG, and the manual e2e drill scripts updated to assert the new hold-until-committed contract.
- Plus `📦 deps(security)`: `@grpc/grpc-js` 1.14.3→1.14.4 and e2e `joi` override →18.2.1 (fresh CVEs flagged by the osv-scanner gate).

## Verification

- Full app gate: 360 files / 10,581 tests, 100% statements/branches/functions/lines.
- Full UI gate: 100% coverage incl. new overlay-hold poll logic and EventSource-close tests.
- `qlty check --all`: no issues. Builds + lints clean across app/ui/e2e.


biggest-littlest (approved):
Reviewed: self-update overlay hold logic, per-operation finalize secret, grace-window reconcile, and SSE close behavior. LGTM.

ALARGECOMPANY (approved):
LGTM — status-poll recovery path and watch-exclusion label verified.

PR #432, opened 2026-06-13
📝 docs(edge): document lookout/1.0 edge endpoint + Ed25519 key registry (M5)

scttbnsn:
## Summary

Documents the **lookout/1.0 edge WebSocket endpoint + Ed25519 agent-key registry** shipped in #429 (M5), which previously had zero docs.

New `content/docs/current/api/lookout.mdx` covers:
- **Ed25519 auth model** — operator registers the agent's public key; agent signs a hello with the canonical `GET\n/api/lookout/ws\n<body-hash>\n<timestamp>\n<nonce>` string; ±60s clock-skew + nonce replay protection; keyId = `hex(SHA-256(raw32)[:8])`
- **Key registry REST API** — `GET`/`POST`/`DELETE /api/lookout/keys` (list/register/revoke) with exact request/response shapes and status codes
- **WebSocket endpoint** — path + `/api/v1` alias, `lookout/1.0` subprotocol, connection flow, protocol limits (30s hello timeout, 16 MB max frame), error frames
- **Agent enrollment** walkthrough

Plus: `lookout` added to the API nav, a cross-link from `agent.mdx`, and a CHANGELOG `### Added` entry.

## Verification

Every API shape was checked against the source (`app/api/lookout.ts`, `app/api/lookout-ws.ts`, `app/store/agent-keys.ts`). Documents **REST enrollment only** — the `authorized_keys` file loader exists in code but is not wired to server startup, so it is intentionally not documented as an operator feature (docs-truth-up discipline). No `apps/web` changes.

PR #430, opened 2026-06-13
✨ feat(agents): declare lookout runtime info fields in ack payload (M4)

scttbnsn:
## Summary

Implements the drydock half of milestone **M4**: declares the `logLevel` and `pollInterval` runtime-info fields in `AgentRuntimeAckPayload` and threads them into the agent info surface. Resolves integration gap **G4** (a symmetric omission — both sides previously dropped these fields).

Paired with lookout PR **`feat/agent-info-fields`**, which populates the same fields on the wire. The contract is locked field-for-field:

| Field | lookout (Go json tag) | drydock (TS) |
|-------|----------------------|--------------|
| `logLevel` | `json:"logLevel"` (string) | `logLevel?: unknown` |
| `pollInterval` | `json:"pollInterval"` (string, seconds) | `pollInterval?: unknown` |
| `memoryGb` | `json:"memoryGb"` (float64, GiB) | `memoryGb?` (already existed) |

`buildRuntimeInfoFromAck` threads both new fields with the same string-guard / non-empty / fallback pattern as sibling fields. UI surfacing reuses the existing detail-field pattern in `AgentsView.vue`.

## Validation

```
all tests pass · 100% line/branch/function/statement coverage maintained
```

A cross-repo pairing review confirmed the JSON field names, types, and units match exactly on both sides (no ms/s mismatch on `pollInterval`; `memoryGb` is GiB on both).

## ⚠️ Merge note

This branch and **`feat/lookout-edge`** (M5) both modify `app/agent/AgentClient.ts` — whichever merges second needs a trivial rebase.

🎯 Target: `main`

PR #429, opened 2026-06-13
✨ feat(edge): lookout/1.0 edge WebSocket endpoint with Ed25519 auth (M5)

scttbnsn:
## Summary

Implements milestone **M5**: the controller-side **`GET /api/lookout/ws`** WebSocket endpoint that lets a lookout agent in **edge mode** (dialing *out* through NAT/firewalls) attach to drydock. This closes integration gap **G1** and makes edge mode functional end-to-end for the first time — lookout's edge client was already complete; drydock was the missing half.

## What's included

- **Ed25519 public-key registry** (`store/agent-keys.ts`, LokiJS-backed) — the controller stores only *public* keys; supports per-agent keys + revocation. Operator add/revoke via REST (`api/lookout.ts`).
- **`/api/lookout/ws` gateway** (`api/lookout-ws.ts`) — verifies a signed `lookout/1.0` hello: keyId lookup → timestamp skew → nonce replay check → `ed25519.verify` over the exact byte construction lookout signs → welcome. Reconstructs SPKI DER from the raw 32-byte key (Node 24 `crypto.verify` doesn't accept `format:'raw'`).
- **`EdgeAgentAdapter`** (`agent/EdgeAgentAdapter.ts`) — adapter-wraps the existing `AgentClient` pipeline (no parallel client), translating `lookout/1.0` frames (container/component sync, exec subprotocol, streaming responses, ping) into the established dispatch path.

## Security hardening (from adversarial review)

An adversarial Ed25519 review + frame-by-frame protocol-conformance review ran against this branch; every confirmed finding is fixed in-branch:

- **Nonce-cache poisoning (high):** nonces are now committed *only after* signature verification — an unauthenticated caller can no longer exhaust the replay pool.
- **Streaming responses dropped (high):** fixed a `pendingRequests` key-prefix mismatch (`stream:` ) that silently no-op'd all streaming Docker tunneling.
- **Exec subprotocol incomplete (high):** added `sendInput` / `sendResize` / `outputHandler` wiring so exec sessions are bidirectional.
- Plus: per-key nonce admission limit, agentId type/length/charset validation, generic (non-enumerable) auth errors, 32-byte key guard, TOCTOU fix on the duplicate-agent guard.

## Validation

```
364 test files · 10,681 tests pass · tsc build clean · biome clean
```

A **live end-to-end test passed**: a real lookout binary completed the signed hello → welcome → container-sync handshake against this endpoint, and an **unregistered key was rejected** (negative control).

## ⚠️ Merge notes

[...]

PR #433, opened 2026-06-13
✨ feat(edge): finalize lookout/1.0 for RC — experimental gate, audit bug fixes, docs (M4+M5)

scttbnsn:
## What & why

A multi-agent audit of the merged **lookout/1.0** edge integration (M4 #430 + M5 #429) found the code fundamentally sound — Ed25519 auth is correct, the SPKI header byte-exact, nonce-commit-after-verify ordering safe, the CodeQL TOCTOU fix genuine — but surfaced confirmed protocol/auth bugs, documentation inaccuracies, and a release-readiness gap: the feature shipped to `main` with **no CHANGELOG entry** and its docs were stranded in the unmerged #432.

This PR makes lookout RC-ready and **supersedes #432** (whose docs had real accuracy bugs, fixed here). Per decision: ship the feature **experimental / opt-in** for this RC, and fix the confirmed code bugs.

### ✨ Experimental gate (default off)
- `DD_EXPERIMENTAL_LOOKOUT=true` is now required to mount the `lookout/1.0` WebSocket endpoint and the `/lookout` key-registry router. Default-off = **zero runtime footprint**, so the endpoint can soak as an opt-in before any GA commitment. The wire protocol may change while experimental.

### 🐛 Audit-confirmed bug fixes (each with tests; 100% coverage held)
- **Stream-request error frames no longer dropped** — `EdgeAgentAdapter` routed error frames under the bare `requestId` while streaming requests are keyed `stream:${requestId}`, so a streaming call would hang the full 30 s timeout instead of rejecting. Now rejects immediately.
- **`startExec` no longer sends an `exec_end` for a never-started `execId`** when the session limit is hit (was a subprotocol violation).
- **Revoked keys in the `authorized_keys` file no longer silently re-activate on restart** — the idempotency check now matches any record (active or revoked) and warns instead of re-inserting.
- **Key revocation now disconnects live sessions** — `DELETE /lookout/keys/:keyId` closes any live WebSocket authenticated under that key (`disconnectByKeyId`).
- **`DELETE /lookout/keys/:keyId` validates the 16-hex keyId format** (400) before use, with a fixed message (no raw input reflection).
- **Hello-timeout timer cleared on premature close/error**; non-finite timestamp rejected; over-long signature rejected before the crypto allocation.

### 📝 Docs + spec
- `lookout.mdx`: flip path hierarchy to `/api/v1/...` primary (`/api/...` deprecated, removal v1.6.0); document the **base64url** signature encoding (vs standard-base64 pubkey); add the hello-frame schema using the real `pubKeyId` field name; add the welcome-frame shape; complete the **13-code** error catalogue; correct the revoke-404 wording; prominent `DD_EXPERIMENTAL_LOOKOUT` callout.
- `agent.mdx`: add the M4 `logLevel`/`pollInterval` fields to the `dd:ack` example.
- **OpenAPI 3.1** path definitions for `GET`/`POST`/`DELETE /api/v1/lookout/keys` (new `Lookout` tag).
- README + ROADMAP: experimental edge-agent dial-out (Ed25519) in features and the v1.5.0 row.
- **CHANGELOG**: new `Added` entries for M5 (PR #429) and M4 (PR #430) — both were previously absent from `main`'s changelog entirely.

## Verification
- **app: 10,834 tests pass, 100% coverage** (lines/branches/functions/statements).
- `tsc` build clean; biome clean (the 2 remaining warnings are pre-existing write-only-field warnings on `EdgeAgentAdapter`, unchanged from `main`).

Supersedes #432.


biggest-littlest (approved):
Approved — consolidated lookout RC-readiness PR. Verified: experimental gate defaults off, audit bug fixes carry tests, docs corrected (paths/base64url/pubKeyId/error codes), CHANGELOG M4+M5 present. Local gate green (100% cov app+ui).

ALARGECOMPANY (approved):
Second code-owner approval. Endpoint is opt-in via DD_EXPERIMENTAL_LOOKOUT; revocation now disconnects live sessions and revoked keys no longer reactivate on restart. LGTM for the RC.

PR #439, opened 2026-06-16
🎨 style(brand): refresh the drydock whale logo everywhere

scttbnsn:
New master render (`drydock.png`, repo root) replaces the whale brand mark across every surface, all regenerated from the single master via the new committed `scripts/regenerate-brand-assets.sh` (ImageMagick).

**Replaced (filenames preserved in place):**
- **App UI** (ships in the image): `drydock-logo.png`, `whale-logo.png`, `favicon.ico`/`.svg`/`-96`, `apple-touch-icon`, PWA manifest icons
- **Website (`apps/web`) + demo (`apps/demo`)**: favicons, PWA icons, OpenGraph cards, header logo
- **README + docs**: `docs/assets/whale-logo.png` and the dark-mode variant (RGB-negate → orange, mirroring the existing `favicon.svg` dark `invert(1)`)

**Untouched:** CodesWhat org logo, `selfhst/*` service icons, app screenshots.

**Safety:** no renames — `docs/assets/whale-logo.png` is the Home Assistant `entity_picture` URL contract (`Hass.ts`); renaming broke HASS once (CHANGELOG #138).

PR #440, opened 2026-06-16
✨ feat(cli): colored startup banner from the whale logo

scttbnsn:
## What

drydock now renders a compact colored banner of the whale logo on startup when attached to an interactive terminal, followed by a `drydock v<version> · <mode>` identity line.

## How

- **Build-time bake.** `scripts/gen-banner.mjs` samples the master logo (`drydock.png`) to a 50-cell-wide grid and renders each terminal row as two pixel rows via the `▀` half-block (upper pixel = foreground, lower = background) using 24-bit truecolor escapes. Transparent cells become spaces; blank top/bottom rows are trimmed. The result is baked into `app/banner/art.ts` so **startup decodes no image** and pulls in no image library. The generator is deterministic — re-running produces no diff.
- **Render.** `renderBanner({ mode })` writes the art + identity line to **stderr**, centering it when the terminal is wider than the banner.
- **Quiet by default where it matters.** Suppressed automatically when the stream is not a TTY or when `NO_COLOR` is set (non-empty), so piped output, log capture, and CI stay clean. Called from the entrypoint for both controller and agent modes.

## Tests

- `app/banner/index.test.ts` — 11 cases covering TTY gating, `NO_COLOR` (set / empty / unset), centering vs no-padding vs undefined columns, the stderr/env fallbacks, and identity-line content for both modes.
- `app/index.test.ts` — asserts `renderBanner` runs with `{ mode: 'controller' }` and `{ mode: 'agent' }`.
- Full app gate green locally: 366 files / 10,875 tests, 100% coverage.

Builds on the new master logo landed in #439.


biggest-littlest (approved):
Clean TTY-gated banner. Art baked at build time so startup pulls in no image decoder, and it's suppressed for non-TTY/NO_COLOR so logs and pipes stay clean. Coverage is full. LGTM.

ALARGECOMPANY (approved):
Generator is deterministic, both controller/agent modes are exercised, and the gate is green. Nice touch. Approving.

PR #441, opened 2026-06-16
🔧 chore(security): drop Snyk, consolidate dep/CVE scanning on Grype

scttbnsn:
Decommissions Snyk and consolidates dependency/CVE scanning on **Grype**.

## Why
Snyk's GitHub SCM integration scans the full dependency *requirement graph* across every `package.json`/`package-lock.json` rather than the resolved, shipped dependency set, so it over-reports advisories in transitive packages the lockfile never resolves to — noise on top of a redundant paid integration that, in practice, was never even wired up (`SNYK_TOKEN` was never set, so the weekly jobs always took the skip path). Grype matches lockfile-resolved versions and scans the built image's package catalog, so no requirement-graph false positives.

## What

**Removed**
- `.snyk` policy file
- `.github/workflows/security-snyk-weekly.yml` (6-job paid-scan workflow)
- `.github/actions/setup-snyk/` composite action
- `scripts/snyk-*` gate + quota scripts (+ its test)
- README / website (`apps/web`) Snyk badges; reworked the CONTRIBUTING scanning section; updated `docs/ci-flow.html`; dropped the stale Snyk cards from `docs/audit-findings.html`

**Added — `security-grype.yml`** (no `govulncheck`; drydock is TypeScript/Node, not Go)
- `grype-deps` — `anchore/scan-action path: .` over all six npm lockfiles (root, app, ui, e2e, apps/demo, apps/web); runs on PRs (path-filtered to deps/Dockerfile/workflow) + weekly cron + dispatch
- `grype-image` — builds + scans the container; `if: github.event_name != 'pull_request'` so PRs don't pay for a Docker build
- SHA-pinned, harden-runner, `persist-credentials: false`, distinct SARIF categories, upload gated on public-repo visibility, `fail-build` on HIGH/CRITICAL

CodeQL (SAST), `dependency-review`, OpenSSF Scorecard, and zizmor remain as the free gates. Trivy intentionally not added.

## Verification
- actionlint + zizmor clean; 27 workflow-invariant test files (93 tests) pass
- `git archive HEAD | grype dir:… --fail-on high` → **No vulnerabilities found**
- Full pre-push gate green (lint, qlty, coverage 100%, build)
- Code-scanning dashboard had zero Snyk analyses/alerts in any state (token was never configured)

The Snyk org-dashboard + GitHub-app teardown is the only remaining step, done by the owner in the Snyk UI — the repo side is clean. The unused `ci-security` environment was already deleted.

PR #442, opened 2026-06-16
🔒 security(image): clear image HIGH/CRITICAL CVEs and scope the Grype gate

scttbnsn:
## What

Clears the HIGH/CRITICAL container-image CVEs that the new `grype-image` scan
(added in #441) surfaced, and scopes the Grype image gate so it stays accurate
going forward.

Before this change the built release image reported **11 critical + 22 high**.
After it: **0 HIGH/CRITICAL outside the two bundled third-party tool binaries**
(`cosign`, `trivy`), which are triaged with a documented, location-scoped
`.grype.yaml`.

## Changes

- **`Dockerfile`** — bump the base + pins that carried the fixable CVEs:
  - `node:24-alpine` → Alpine 3.24.1 digest (clears the Node-runtime CVE, musl, curl ×2, git)
  - `bash 5.3.3-r1 → 5.3.9-r1`, `git 2.52.0-r0 → 2.54.0-r0`
  - `cosign 2.6.3-r1 → 3.0.6-r1` (clears the bulk of the vendored-Go cosign CVEs; CLI `verify` flags drydock uses are unchanged)
- **`.grype.yaml`** (new) — location-scoped `ignore` for `/usr/bin/cosign` and
  `/usr/bin/trivy` only. These are vendored Go-module CVEs compiled into the
  upstream tool binaries; we can't patch them independently — they clear when
  Alpine rebuilds those packages, which we pick up on the next base bump. The
  Node runtime, every OS package, and the entire app dependency graph stay
  fully gated. Gating on a module graph we don't ship is the same
  manifest-vs-shipped mismatch we dropped Snyk over.
- **`security-grype.yml`** — wire `GRYPE_CONFIG` to the repo-root `.grype.yaml`
  so the gate is deterministic across action versions (grype also auto-discovers
  it, but pinning the path is explicit).
- **`CHANGELOG.md`** — `Security` entry under Unreleased.

## Verification

- Full release image scan: `grype --fail-on high` with the config exits **0**
  (was exit 2 / 33 findings). HIGH+ findings outside the two triaged binaries: **0**.
- cosign 3.0.6 CLI confirmed compatible with `app/security/scan.ts`'s
  `verify --output json --key --certificate-identity --certificate-oidc-issuer`.
- qlty / hadolint / yamllint clean on changed files.

The red `grype-image` alerts currently in the Security tab will auto-close on the
next scan once this lands (the new SARIF no longer reports them).

### DIFF
--- a/.github/workflows/security-grype.yml
+++ b/.github/workflows/security-grype.yml
@@ -125,6 +125,12 @@ jobs:
         # shipped — so this is the authoritative view of the runtime CVE surface.
         uses: anchore/scan-action@e1165082ffb1fe366ebaf02d8526e7c4989ea9d2  # v7.4.0
         id: grype
+        env:
+          # Load the repo-root .grype.yaml explicitly (it documents why the
+          # bundled cosign/trivy CLI binaries' vendored Go-module CVEs are
+          # triaged). grype also auto-discovers it from the workspace, but
+          # pinning the path keeps the gate deterministic across action versions.
+          GRYPE_CONFIG: ${{ github.workspace }}/.grype.yaml
         with:
           image: drydock:grype-scan
           severity-cutoff: high
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -1,0 +1,28 @@
+# Grype configuration — drydock container image scan triage.
+#
+# The grype-image job (.github/workflows/security-grype.yml) fails the build on
+# HIGH/CRITICAL. These ignore rules scope that gate to the dependencies drydock
+# actually controls — the Node runtime, the Alpine OS packages, and the app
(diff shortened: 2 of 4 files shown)


biggest-littlest (approved):
Image CVE remediation looks right. Base + cosign bump clears the fixable HIGH/CRITICALs; the .grype.yaml triage is location-scoped to the bundled cosign/trivy binaries only and well documented. Verified the gate passes. 👍

ALARGECOMPANY (approved):
LGTM. Triage rationale matches the Snyk-removal reasoning (don't gate on a module graph we don't ship), Node/OS/app deps stay fully gated. Approving.

PR #445, opened 2026-06-22
🧹 chore: pre-1.5.0 cleanups — EdgeAgentAdapter dead code + Crowdin HOME fix

scttbnsn:
Two small, unrelated pre-GA cleanups (split into two commits).

## 🗑️ Remove dead code from `EdgeAgentAdapter`
The adapter stored `this.hello` and `this.adapterConfig` from its constructor but never read either — biome flagged both as unused private members (warn-level). The gateway (`portwing-ws.ts`) already applies `hello.version` to `client.info` before constructing the adapter, and the agentId is encoded in the client name, so these were pure dead copies.

Removed the two fields, the two constructor params, and the now-orphaned `EdgeAgentAdapterConfig` interface; updated the single call site (`portwing-ws.ts`) and the test. `HelloMessage` stays — `portwing-ws.ts` still parses the hello frame with it. Verified: biome clean, `EdgeAgentAdapter.test.ts` (84 tests) green, `tsc` clean.

## 🐛 Fix Crowdin sync gitconfig permission error
`🌐 i18n: Crowdin Sync` has failed every run since **2026-05-28**. The `crowdin/github-action` container inherited `HOME=/` from the runner, so git couldn't write `//.gitconfig` (`could not lock config file ... Permission denied`) — translations pulled fine but the commit/PR step always failed. Pinned `HOME` to the mounted, writable `/github/home` in the step env.

Verified by dispatching the workflow on this branch before merge (the `crowdin` environment has no branch restriction).

### DIFF
--- a/.github/workflows/i18n-crowdin.yml
+++ b/.github/workflows/i18n-crowdin.yml
@@ -48,6 +48,15 @@ jobs:
           pull_request_base_branch_name: main
           commit_message: "🔧 chore(i18n): sync translations from Crowdin"
         env:
+          # The crowdin/github-action entrypoint runs `git config --global`
+          # (safe.directory, then the commit user) but the container inherits
+          # HOME=/ from the runner, so those writes hit "could not lock config
+          # file //.gitconfig: Permission denied" and the commit/PR step fails.
+          # A step-level HOME can't fix it: GitHub passes HOME to container
+          # actions via `-e HOME` pass-through, which always wins. Instead point
+          # git's global config at a mounted, writable file — GIT_CONFIG_GLOBAL
+          # is honored independent of HOME (git >= 2.32).
+          GIT_CONFIG_GLOBAL: /github/home/.gitconfig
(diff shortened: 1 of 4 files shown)


biggest-littlest (dismissed):
Two clean pre-GA cleanups. Edge dead-code removal is correct — gateway already sets client.info from hello, so the stored copies were dead. Crowdin HOME pin is the right fix for the //.gitconfig perm error. LGTM.

ALARGECOMPANY (dismissed):
Verified the edge refactor keeps HelloMessage (still used by the gateway) and the test passes. Crowdin fix verified by branch dispatch. Good to go.

biggest-littlest (approved):
Re-approving after the GIT_CONFIG_GLOBAL correction. Verified root cause via the action entrypoint — all git config --global writes now target a writable path.

ALARGECOMPANY (approved):
Re-approve: GIT_CONFIG_GLOBAL is the correct lever (HOME can't be overridden for container actions). Pending branch-dispatch verification.

PR #448, opened 2026-06-23
🐛 fix(portwing): complete edge wire-contract integration with portwing agent

scttbnsn:
## What

Completes drydock's side of the edge wire-contract so it lines up byte-for-byte with the portwing edge agent. These are the controller-side halves of the integration audit (portwing carries the matching agent-side changes in CodesWhat/portwing#31).

## Changes

- **O3 — disabled-endpoint log:** when `DD_EXPERIMENTAL_PORTWING` is off, log that the `portwing/1.0` edge endpoint is disabled instead of silently 404ing.
- **O4 — stream inactivity timeout:** the request/stream timer now resets on each chunk (inactivity timeout) rather than a single fixed deadline, so long-lived streams aren't killed mid-flight.
- **O5 — exec_end on teardown:** `session.close()` sends an `exec_end` frame to the edge before dropping the session; `onDisconnect` tears down every live session.
- **O7 — full metrics surface:** `handleMetrics` reads the 9 additional runtime fields (cpu/mem/disk/network) the agent now reports.
- **O9 — streamed responses:** a `response` frame with `isStream=true` stashes statusCode/headers and defers resolution to `stream_end`.
- **O10 — chunk assembly:** stream chunks are base64-decoded and concatenated into the response body.
- **O11 — authorized keys file:** load `DD_PORTWING_AUTHORIZED_KEYS` at startup so edge agents don't need manual key registration.
- **O12 — exec tty:** `startExec` carries a `tty` option (default true) end to end.

## Tests / coverage

drydock enforces 100% coverage (lines/branches/functions/statements). All new code is covered — added cases for the exec_end teardown path, the stream inactivity timeout, the metrics fields, the isStream response/chunk assembly, the authorized-keys load, and the false-guard branches on the stream-response path. Full gate green: `21218/21218` statements, `13391/13391` branches.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/agent/AgentClient.ts
+++ b/app/agent/AgentClient.ts
@@ -67,6 +67,15 @@ interface AgentClientRuntimeInfo {
   lastSeen?: string;
   logLevel?: string;
   pollInterval?: string;
+  cpuUsage?: number;
+  cpuCores?: number;
+  memoryUsed?: number;
+  memoryFree?: number;
+  diskTotal?: number;
+  diskUsed?: number;
+  diskFree?: number;
+  networkRxBytes?: number;
+  networkTxBytes?: number;
 }
 
 interface AgentComponentDescriptor {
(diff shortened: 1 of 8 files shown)


biggest-littlest (approved):
Controller side lines up with the portwing agent wire contract — stream/exec_end and authorized-keys paths read right. Coverage gate's green. LGTM.

ALARGECOMPANY (approved):
Went through the isStream response + chunk assembly and the exec tty default. Looks correct. Approving.

Issue #453, opened 2026-06-26
Feature request: fetch and display all intermediate release notes between current and target version


steilerDev:
## Summary

Release-notes enrichment currently fetches notes for exactly two versions — the **current** tag and the **update-target** tag — and shows them side by side. When a container is several versions behind, there is no way to see what changed in the **intermediate** releases. That gap matters: deciding whether to update often hinges on a breaking change, migration step, or security fix introduced in a version *between* current and latest, which is invisible today.

Please consider fetching and displaying **all releases between the current and target version**, so operators can make an informed update decision.

## Current behavior (v1.5.0)

`enrichContainerWithReleaseNotes` fetches only two single-tag releases:

- `app/watchers/providers/docker/release-notes-enrichment.ts:27` — `currentReleaseNotes` for the current tag
- `app/watchers/providers/docker/release-notes-enrichment.ts:43` — `releaseNotes` for the update-target tag (`container.result.tag`)

Each is a single-release lookup:

- `app/release-notes/providers/GithubProvider.ts:193` — `GET /repos/{owner}/{repo}/releases/tags/{tag}`

The UI renders the two side by side (`ui/src/components/containers/ReleaseNotesLink.vue`, container fields `releaseNotes` / `currentReleaseNotes`). Nothing aggregates the releases in between, and if current and target resolve to the same release the second fetch is skipped (`release-notes-enrichment.ts:38-41`).

## Requested behavior

When an update is available and current/target tags are both semver-resolvable:

1. Enumerate the repo's releases between current (exclusive) and target (inclusive) — e.g. via the paginated `GET /repos/{owner}/{repo}/releases` list endpoint — and order them by semver.
2. Surface them as an ordered list (newest→oldest or oldest→newest) in the release-notes popover, in addition to (or merged with) the current/target panels that exist today.
3. Keep it best-effort: fall back gracefully to today's two-panel behavior when intermediate releases can't be determined.

## Technical considerations / nuances to call out

- **Ordering requires semver.** Intermediate selection only works when tags are semver-comparable (and after any `transformTags` is applied). Non-semver / date / rolling tags can't be ordered into a range — fall back to the current two-version view for those.
- **Tag ↔ release matching.** The existing `buildTagVariants` logic (prefix `v`, `.git` handling, etc.) must be reused so listed releases map back to the watched tags correctly. Some repos also publish tags without a corresponding GitHub *release* — those would have no notes and should be skipped or shown as "no release notes."
[...]

PR #459, opened 2026-06-28
🚀 release(v1.5.1-rc.3): intermediate release notes, i18n re-sync, web shell + renovate

### LINKED ISSUE #452: Allow DD_RELEASE_NOTES_GITHUB_TOKEN to override source-repo "trust" for release-notes lookups (and document narrow scoping)
## Summary

When the release-notes source repo is resolved from an **untrusted** source (an operator-set `dd.source.repo` *container* label, a persisted `container.sourceRepo` fallback, or a live Docker Hub metadata lookup), the GitHub auth token is unconditionally suppressed — **including the dedicated `DD_RELEASE_NOTES_GITHUB_TOKEN`**, not just the GHCR fallback. There is no configuration or env override to opt back in.

This makes authenticated release-notes lookups impossible for the common case of a **third-party Docker Hub image that lacks OCI source labels**, where the operator can only supply the repo via a container label. The operator is permanently capped at the unauthenticated GitHub API limit (60 req/h, public repos only), even when they have deliberately created a narrowly-scoped, read-only-public token whose leakage would be harmless.

## Current behavior (v1.5.0)

The `trusted` flag is purely structural — decided by *where* the repo string came from — and short-circuits at the container label:

- `app/release-notes/index.ts:223-228` — a container-level `dd.source.repo` returns `{ trusted: false }` **immediately**, with no comparison against the image's OCI labels.
- `app/release-notes/index.ts:352-354` — `trusted` is passed straight through as `allowToken`:
  ```ts
  const releaseNotes = await provider.fetchByTag(sourceRepo, tag, getGithubToken(), {
    allowToken: trusted,
  });
  ```
- `app/release-notes/providers/GithubProvider.ts:186-190` — when `allowToken === false`, **both** the explicit `DD_RELEASE_NOTES_GITHUB_TOKEN` and the GHCR fallback are dropped:
  ```ts
  const allowToken = options?.allowToken !== false;
  const effectiveToken = allowToken ? (token ?? getGhcrTokenFallback()) : undefined;
  ```

The behavior is documented as intentional in `content/docs/current/configuration/watchers/index.mdx:757`.

## Why this is a problem

The threat model — a malicious container spoofing `dd.source.repo` to exfiltrate a high-scope token to an attacker-controlled repo — is legitimate for the **GHCR fallback token**, which the operator provisioned for a different purpose and may be broadly scoped. But it should not force the same restriction onto a token the operator created *specifically* for release notes. An operator who scopes that token to read-only public access has explicitly accepted the (negligible) risk, and there is currently no way to express that intent.

## Proposed change

Treat the presence of a **dedicated `DD_RELEASE_NOTES_GITHUB_TOKEN`** as an explicit operator opt-in to use that token regardless of source trust:

- When `DD_RELEASE_NOTES_GITHUB_TOKEN` is set, send it even for untrusted (container-label / Docker Hub / persisted) sources.
- Keep the GHCR fallback (`getGhcrTokenFallback()`) gated by `trusted` as it is today — only the explicitly-configured release-notes token gets the override. This preserves the original protection for the broadly-scoped GHCR credential while honoring deliberate operator intent.

(Alternatively, a separate explicit flag such as `DD_RELEASE_NOTES_TRUST_CONTAINER_LABELS=true` would also solve it, but overloading the dedicated token is cleaner — configuring it is already the explicit signal.)

## Docs request

Please document this feature explicitly in the release-notes / GHCR configuration docs, including **precise instructions to narrowly scope the token** so operators can opt in safely. Concretely:

- A **classic PAT** scoped to **`public_repo`** only (no other scopes), or
- A **fine-grained PAT** with read-only **Contents** access limited to **public repositories** and **no account/repo write permissions**.

Make clear that this token — unlike the GHCR fallback — is sent even for operator-supplied `dd.source.repo` container-label sources, and that it should therefore be created exclusively for read-only public release-notes access.

## Secondary issue (related, worth a doc note regardless)

Because the container-label check short-circuits before the OCI labels (`index.ts:223-228`), adding a `dd.source.repo` **container** label to an image that *already* carries a trusted `org.opencontainers.image.source` label **silently downgrades** the resolution from trusted to untrusted and drops the token. This is surprising and currently undocumented — at minimum it deserves a warning in the `dd.source.repo` docs ("don't set this as a container label if the image already exposes a trusted OCI source label").

## Environment

- Version: 1.5.0

scttbnsn:
This RC consolidates four streams of held work plus the docs/build fixes they surfaced. One branch, one PR, one cut.

## Release notes — intermediate range (#452, #453)

- New `GET /api/containers/{id}/intermediate-release-notes` endpoint, lazy-loaded when the popover opens — not embedded in the container model or agent snapshot, so it adds no ongoing payload weight. Accepts `from` (required) and `to` (defaults to the pending update tag).
- The release-notes popover now shows the full range between the running and target version. Best-effort, semver-only; date/rolling tags (`latest`, `stable`) fall back to the two-panel current/target view. Cap via `DD_RELEASE_NOTES_MAX_INTERMEDIATE` (default `20`, `0` disables); `__FILE` secret-file convention supported. Over the cap, the popover shows a non-silent "N older releases not shown" notice.
- `DD_RELEASE_NOTES_GITHUB_TOKEN` is now forwarded to repos resolved from a `dd.source.repo` container label or persisted `container.sourceRepo`. The GHCR token fallback stays restricted to trusted sources (OCI labels, GHCR image paths) and is never sent to a container-label source. A `warn` line fires each watch cycle when a `dd.source.repo` label shadows a trusted OCI source label.
- Three-tier cache with a hardened intermediate-range caching contract.

## i18n / Crowdin re-sync (supersedes #458)

- Re-synced the 16 target-locale `containerComponents.json` catalogs from Crowdin so key order tracks the English source.
- Guard test asserting on-disk locale dirs match `SUPPORTED_LOCALES`.

## CodesWhat web shell — apps/web

- fumadocs v15 (mdx 15, core/ui 16.10), Next 16.2.9, React 19.2.7, Tailwind 4.3.1.
- Moved app/lib/components into `src/` to match the shared shell architecture; promoted lab section components to flat `src/components`.
- `src/lib/site-config.ts` is the single reskin point; all brand couplings wired to it. Comparison scaffold made product-agnostic.
- Whale OG share image (1200×630) + regenerated favicon suite from the new headshot; config-driven version badge, manifest icon purpose + theme_color.
- a11y fixes for site chrome, dead-code purge, `sync-docs` made portable + atomic (stops mutating the source tree).
- Block-formatted the release-notes Callout in the watchers docs so fumadocs v15 builds.

## Renovate (supersedes #460)

- Adopt the shared CodesWhat Renovate preset.

## Gate

Full pre-push gate green: biome, qlty, qlty-smells, coverage (app + ui, 100%), build, e2e, playwright, zizmor. apps/web build verified green through the real `sync:docs` path.

Closes #452, #453. Supersedes #458 and #460 (closed in favor of this consolidated RC).


biggest-littlest (dismissed):
RC consolidation looks clean — intermediate release-notes endpoint + UI, the Crowdin re-sync with the locale guard test, the fumadocs v15 web-shell move, and the renovate preset all accounted for, full gate green. LGTM.

ALARGECOMPANY (dismissed):
Went through the four streams and the build/coverage/e2e gate. Scope matches the changelog, nothing dangling. Good to merge.

biggest-littlest (approved):
Re-approving after the CodeQL sanitization fix — encodeURIComponent on the repo slug clears the incomplete-sanitization alert, badge URL unchanged for the real owner/repo. All required checks green.

ALARGECOMPANY (approved):
Re-approve post-fix. CodeQL clean, full gate green, scope unchanged.

PR #468, opened 2026-07-02
🐛 fix(ui): shrink table columns to fit and pin the group Update All button (#467)

### LINKED ISSUE #467: Update All buttons in Container>Stack View no longer rendering
### Drydock version

1.5.1-rc.4

### What happened?

See attached

Image

### What did you expect?

See above

### Steps to reproduce

See above

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

24.0.1

### OS / Architecture

Synology DSM

scttbnsn:
Fixes #467 (v1.5.1-rc.4 regression, GA-blocking).

## Root cause

The rc.4 v6→v7 preferences migration force-adds the new **Version** column (~220px) to every existing user's containers column set. The responsive auto-hide budget sums column **minimum** widths, but the renderer laid columns out at their **preferred** widths and never shrank them. In the band between those two sums (~1324–1544px of content width for the default column set) the table overflows horizontally, and the group-header **Update All** button — a plain flex child of a full-row `<td colspan>`, unlike the `sticky right-0` per-row Actions cells — scrolls out of the visible area. Badge and row action icons stay visible, the button silently disappears. Root cause was established by a 5-agent triage workflow and upheld by two independent adversarial verifiers against the rc.3/rc.4 sources.

## Fix

- `DataTable.resolveColumnWidths()` now shrinks columns proportionally toward (never below) `minSize` when base widths overflow the available width (water-filling with per-pass clamping). Rendered widths now agree with the auto-hide budget's minSize-based "fits" decision, eliminating the divergence band. Expansion behavior for `extra > 0` is unchanged and pinned by a hand-verified test.
- The group-header Update All button is pinned to the visible end edge (`sticky end-0`, logical props for RTL) so it stays reachable even when a table legitimately overflows between auto-hide steps.
- CHANGELOG: `[1.5.1-rc.5]` section added.

## Testing

- New regression spec reproduces the exact rc.4 post-migration column set in the overflow band (available=1300 between minSize-sum 1136 and size-sum 1484) and asserts rendered widths fit with every column ≥ its minSize.
- Floor spec: below the minSize-sum, columns render at exactly minSize and residual overflow is left to auto-hide.
- Expansion-unchanged spec with exact arithmetic; sticky-wrapper spec on the group header.
- Full gate: 3760 ui tests, 100% coverage on all four metrics (app suite also green in pre-push), typecheck, vite build, biome, qlty all clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.5.1-rc.5] — 2026-07-02
+
+### Fixed
+
+- **Grouped "Update All" buttons could scroll out of view after upgrading to rc.4.** The new Version column added in rc.4 widened the containers table for existing users, and at moderate desktop widths the table overflowed horizontally in a way its responsive column-hiding never accounted for: columns were only ever laid out at their preferred widths, so the per-stack **Update All** button — positioned at the far end of the group header row — ended up past the visible edge while everything else looked normal. Tables now shrink columns proportionally toward their minimum widths when space is tight (matching the widths the responsive logic already budgets for), and the group header's Update All button is additionally pinned to the visible edge, so it stays reachable even when a table legitimately overflows. (#467)
+
 ## [1.5.1-rc.4] — 2026-06-29
 
 ### Added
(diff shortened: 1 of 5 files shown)


biggest-littlest (approved):
Shrink-to-minSize water-filling is the right class fix — render now agrees with the auto-hide budget, and the sticky pin covers the residual overflow band. Regression spec reproduces the reported width band exactly. LGTM.

ALARGECOMPANY (approved):
Verified the expansion path is pinned unchanged and coverage holds at 100%. CHANGELOG entry reads well. Approving.

PR #474, opened 2026-07-05
🔧 config(coderabbit): add review config with per-path agent rules

scttbnsn:
Adds `.coderabbit.yaml` based on the SYP dev-toolkit template: chill profile, no poem, learnings scoped local, generated/vendored paths filtered.

Per-path rules: registries and triggers get checked against the shared provider pattern (plus secret handling in trigger logs), API handlers don't get flagged for auth that middleware already enforces, tests and docs content get a lighter touch.

Inert until the CodeRabbit app is installed on the org. Public repos get Pro features free, no license requirement.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,63 @@
+# CodeRabbit configuration — drydock
+language: en-US
+
+reviews:
+  profile: chill
+  auto_review:
+    enabled: true
+  high_level_summary: true
+  collapse_walkthrough: false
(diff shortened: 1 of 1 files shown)


biggest-littlest (approved):
Config-only change, reviewed the yaml against the CodeRabbit v2 schema.

ALARGECOMPANY (approved):
Config-only change, reviewed the yaml against the CodeRabbit v2 schema.

PR #481, opened 2026-07-06
🐛 fix(ui): make click-to-copy work on insecure contexts and fix tooltip feedback (#472)

### LINKED ISSUE #472: "Click to copy" doesn't work
### Drydock version

 v1.5.1-rc.4

### What happened?

Hello.

When try to copy digest, it doesn't work.

Image

Image

### What did you expect?

Copy data

### Steps to reproduce

1. Open Drydock
2. Go to specific docker page info.
3. Click "click to copy" anywhere.
4. Show JS error on console 

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

_No response_

### OS / Architecture

_No response_

scttbnsn:
Fixes #472.

Click-to-copy did nothing (and logged a TypeError) on deployments served over plain HTTP, which is the common self-hosted LAN setup, because `navigator.clipboard` only exists in secure contexts. Three commits:

- **execCommand fallback** — `useClipboard` now falls back to the legacy `document.execCommand('copy')` textarea technique when the Clipboard API is missing or rejects. The log viewer's Copy button goes through the same path, and when no copy mechanism works at all the UI shows an explicit "Copy failed" state instead of failing silently. The temp textarea teardown is wrapped in try/finally with a parentNode guard so a mid-copy DOM error can't leave it behind or skip focus restore.
- **Live tooltip updates** — an open tooltip whose text changed (like the "Copied" confirmation) stayed stuck on the old text until you moved the mouse away and back, and the click itself hid the tooltip so the feedback never appeared. The tooltip directive now updates the shared tooltip element in place when its content changes, repositions it for the new text, and re-shows immediately if the pointer never left the anchor.
- **De-duplicated tooltips on CopyableTag** — three call sites (the containers table digest-delta cell and two spots in the dashboard recent-updates widget) bound a second `v-tooltip` directly on `<CopyableTag>`, which falls through to the same root element as the component's own copy-state tooltip and clobbered its directive state. Those sites now pass their text through a new `idleTooltip` prop so one tooltip carries both the informative text and the copy feedback. Swept every `CopyableTag` usage in `ui/src` — no other double bindings.

Gate: full UI suite green at 100% coverage, build/typecheck/biome clean, e2e + playwright via pre-push hook. Manually QA'd on a LAN IP (insecure context) to exercise the fallback path.

🐛 Fixed
- Added a clipboard fallback using `document.execCommand('copy')` so click-to-copy works when `navigator.clipboard` is missing or rejects (e.g., insecure/plain HTTP contexts).
- Made `AppLogViewer`’s toolbar copy button use the same copy helper and surface explicit failure UI (“Copy failed”) when no copy method succeeds.
- Improved copy feedback timing: clears any pending copy-feedback reset timer before scheduling a new one to prevent rapid re-copy indicator glitches.

🔧 Changed
- `useClipboard()` now tracks both success (`isCopied`) and failure (`isFailed`) via `copiedKey`/`failedKey`, and `copyToClipboard()` returns `Promise<boolean>`.
- Introduced exported `writeToClipboard(text): Promise<boolean>` that tries `navigator.clipboard.writeText` first, then falls back to the `execCommand('copy')` textarea path with hardened cleanup + focus restoration.
- Tooltip directive now live-updates while open: when bound text changes it remeasures/repositions, and if the pointer is still over the anchor it re-shows immediately (bypassing the initial delay).
- `CopyableTag` now supports an optional `idleTooltip` prop to prevent duplicate/staked tooltip bindings and preserve correct copy-state tooltip precedence.
- Updated affected `CopyableTag` call sites to pass `idleTooltip` for digest/version displays.
- Added i18n strings for `sharedComponents.copyableTag.clickToCopy` and `sharedComponents.copyableTag.failed`.

✨ Added
- New/expanded tests for:
  - clipboard fallback success/failure, including missing `navigator.clipboard` and failing/throwing `document.execCommand`
  - DOM cleanup edge cases in the fallback path (append/select/throw scenarios)
  - rapid double-copy behavior and correct reset timing
  - tooltip directive dynamic update/re-show behavior
  - `CopyableTag` idleTooltip precedence and failure styling regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **"Click to copy" did nothing and logged a TypeError on deployments served over plain HTTP** (the common self-hosted LAN setup), because the browser Clipboard API only exists in secure contexts. Copying now falls back to the legacy execCommand technique when the API is missing or rejects, covers the log viewer's Copy button too, and shows a "Copy failed" state instead of failing silently when no copy mechanism works at all. (#472)
+
 ## [1.5.1-rc.5] — 2026-07-02
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **"Click to copy" did nothing and logged a TypeError on deployments served over plain HTTP** (the common self-hosted LAN setup), because the browser Clipboard API only exists in secure contexts. Copying now falls back to the legacy execCommand technique when the API is missing or rejects, covers the log viewer's Copy button too, and shows a "Copy failed" state instead of failing silently when no copy mechanism works at all. (#472)
+- **An open tooltip whose text changed — like the copy button's "Copied" confirmation — stayed stuck on the old text until you moved the mouse away and back.** Clicking to copy hid the tooltip outright, so the "Copied"/"Copy failed" state never appeared until a re-hover. Tooltips now update their text in place while open, reposition themselves for the new content, and reappear immediately if the pointer never left. (#472)
 
 ## [1.5.1-rc.5] — 2026-07-02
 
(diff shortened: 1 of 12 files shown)


scttbnsn:
Fixed the reset-timer race in AppLogViewer and the un-awaited clipboard reset in the spec (05c4e3e3). Skipping the i18n nitpick on the digest tooltip: it's the same raw-literal pattern this view already uses (line 675 and friends), and the UI literal sweep is queued with the v1.6 i18n gate work rather than piecemeal here.

biggest-littlest (approved):
LGTM. execCommand fallback is properly guarded with try/finally, tooltip live-update covers the click path, and the double-binding cleanup closes the loop.

ALARGECOMPANY (approved):
Approving — full gate green at 100% coverage, QA'd on an insecure LAN context.

PR #482, opened 2026-07-06
🐛 fix(tag): stop suggesting prerelease tags that coerce to stable versions (#473)

scttbnsn:
Fixes the suggested-tag mislabel from discussion #473.

A Home Assistant `latest` container with a digest update showed the digest correctly but labeled it with an unrelated dev tag (`2026.8.0.dev202607050315`). Root cause: `suggest()` accepted any tag that `semver.coerce()` could mangle into a version — PEP 440-style tags like `2026.8.0.dev202607050315` coerce to a "stable" `2026.8.0`, sail past the prerelease check, and win the sort. The old `PRERELEASE_LABEL_PATTERN` guard missed them because PEP 440 glues the `dev` keyword straight onto the digits with no separator.

The fix replaces the pattern denylist with a structural guard: if a candidate tag only parses via the lossy `coerce()` fallback, it's rejected unless the raw tag is a bare numeric version (`v?X`, `v?X.Y`, `v?X.Y.Z`). That keeps legitimate 1- and 2-part stable tags (`13.4`, `v2`) suggestible while shutting out every keyword-glued prerelease format, not just a hard-coded list.

UI side: the grouped containers view rendered the suggested tag as a raw span; it now uses `SuggestedTagBadge` like the other views, so suggested tags get consistent treatment there.

Gate: app + UI suites green at 100% coverage, build/typecheck/biome clean, e2e + playwright via pre-push hook. Verified against the real registry data from the report (Docker Hub + ghcr digest tables are byte-identical — the digest shown was always correct; only the label was wrong).

🐛 Fixed
- Corrected suggested-tag selection so PEP 440 “dev”/nightly-style prereleases (e.g., `2026.8.0.dev...`) no longer get mis-ranked ahead of the real stable tag (discussion `#473`).
- Rejects tag candidates that only qualify through `semver.coerce()`’s lossy fallback (`requiredCoercionFallback`), unless the raw tag matches a bare numeric form (`v?X`, `v?X.Y`, or `v?X.Y.Z`).
- Adds an invariant documenting why the `normalizeNumericMultiSegmentTag` branch does not need a mirrored guard in `requiredCoercionFallback`.
- Updated grouped container “Update state” rendering so the raw suggested value is shown only via `SuggestedTagBadge`, not as unlabeled text.

🔧 Changed
- Refactored `app/tag/suggest` stability eligibility: removed the prerelease denylist approach and replaced it with structural checks tied to whether coercion fallback was required, while still requiring `parse()` to produce no prerelease parts.
- Updated `ui/src/components/containers/ContainersGroupedViews.vue` to render suggested tags using `SuggestedTagBadge` in the “Update state” cell.

✨ Added
- Expanded `app/tag/suggest.test.ts` with regression/edge-case coverage for `#473` and additional tag-format handling (rejections for dev/post-release suffixes, CalVer-like formats, coercion-lossy OS-variants; acceptances for bare numeric formats and stable build metadata; and semver tie-breaking behavior).
- Added `app/watchers/providers/docker/image-comparison.test.ts` coverage ensuring the digest-watch flow proposes the correct stable tag alongside a PEP 440 nightly/dev tag without disrupting digest comparison logic.
- Updated `ui/tests/components/containers/ContainersGroupedViews.spec.ts` to:
  - Assert no `suggested-tag-badge` renders when fixtures have no `suggestedTag`.
  - Add a `#473` regression validating the badge presence and that the raw suggested tag string is not shown as plain text.


scttbnsn:
Added the invariant comment on requiredCoercionFallback explaining why the normalizeNumericMultiSegmentTag branch doesn't need mirroring here (any tag it rewrites comes back with a non-empty prerelease array and dies at the prerelease check first).

biggest-littlest (dismissed):
LGTM. Round-trip coercion guard is the right shape — keeps bare 2-part stables while killing keyword-glued prereleases.

ALARGECOMPANY (dismissed):
Approving — verified against the reporter's registry data, digest behavior unchanged.

biggest-littlest (approved):
Re-approving after the rebase onto the merged #481 — CHANGELOG resolution only, code unchanged.

ALARGECOMPANY (approved):
Re-approving post-rebase, gate green.

PR #501, opened 2026-07-11
✨ feat(watcher): pinned tags show newer versions as information by default (#498)

### LINKED ISSUE #498: Alert: Pinned tag "1.5.2-rc.1" is compared by digest only. Set dd.tag.family=loose or add a dd.tag.include filter to allow semver version climbing.
### Drydock version

1.5.2-rc.1

### What happened?

Since updating to the latest version (> 1.5.1) I have this warning everywhere:

Pinned tag "1.5.2-rc.1" is compared by digest only. Set dd.tag.family=loose or add a dd.tag.include filter to allow semver version climbing.

Before everything worked better, I saw that some of my containers where old. Now I dont see anything anymore and dont understand it.

Thanks!

Image

Image

### What did you expect?

Correct current version vs most revent version view

### Steps to reproduce

Just run drydock in 1.5.2-rc.1 via Docker

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

_No response_

### OS / Architecture

_No response_

scttbnsn:
Fixes #498 (reopened).

## What

Pinned specific tags (`n8n:2.15.1`, `immich:v2.7.5`) get their version visibility back, by default, with zero config: a new **informational** `updateInsight` field shows the best newer same-family tag (cross-major included) on every pinned container. It is structurally incapable of firing triggers — insight only exists in the pin-gate branch where candidates are hard-empty, so `updateAvailable` stays false and actions never see it. The #321 guarantee (no surprise auto-applied version jumps on pinned tags) is untouched.

Opt-out: `DD_WATCHER_{name}_TAG_PIN_INFO=false`.

## Why

v1.5.0-rc.36 (#321) deliberately stopped pinned tags from producing actionable version candidates, but that also suppressed the *information* that a newer version exists — the core value of a monitor-first dashboard for users running pinned fleets. rc.2 restored digest detection; this restores version visibility, without re-arming the action path.

## Bug fixes riding along

- `dd.tag.family=loose` no longer drops the variant-suffix guard along with the pin gate: a `v2.7.5-openvino` pin can be offered `v3.0.2-openvino`, never bare `v3.0.2` (behavior change for loose users, called out in the CHANGELOG).
- Version sort now prefers same-suffix-template candidates, so a bare tag can't outrank the matching variant.
- Prerelease pins see their own GA: `1.5.2-rc.1` now surfaces `1.5.2` as insight (conservative prerelease-identifier rule; build variants like `-openvino`/`-alpine` keep strict boundaries).
- Same-core suffix-precision growth (`1.2.3-alpine` vs `1.2.3-alpine3.21`) no longer fabricates a phantom patch update.

## Docs

CHANGELOG `[1.5.2-rc.3]`, watchers page (TAG_PIN_INFO, corrected pinned-tag note, 4 previously-undocumented live imgset keys), FAQ (imgset env route for `tag.family`).

## Verification

TDD throughout (new tests proven red first). Both suites at 100% coverage (app 11223 tests, ui 3812). Full pre-push gate green. The diff also went through a 6-dimension adversarial alignment review (trigger-safety path construction, matching edge cases, docs-vs-code, config env round-trip, UI conformance, release-story coherence) — 12 confirmed findings, all fixed in the review-fix commits. Key regression test: a validated container with `updateInsight` populated, driven through the real trigger dispatch at threshold `all`, fires nothing.

## Changelog

- ✨ Added non-actionable informational `updateInsight` (major/minor/patch) for pinned Docker tag pins, including API schema + UI mapping + localization.
- ✨ Added `DD_WATCHER_{name}_TAG_PIN_INFO` (default `true`) to enable/disable pinned tag informational insight computation (non-actionable; never triggers updates).
- ✨ Added `UpdateInsightBadge` and rendered it across container views (table/cards/list/side/full-page), including memoization updates so badge refreshes when insight changes.
- 🔧 Improved tag-family/variant matching and candidate selection: prerelease suffix detection, stricter loose-family suffix/variant guards, and precision-aware semver sorting with exact-suffix tie-breaking.
- 🔧 Restricted prerelease→GA widening to the informational insight path only (via `allowPrereleaseToGA`), keeping actionable update candidates unchanged.
- 🔧 Updated Docker version-finding plumbing to optionally compute insight (`pinInfoEnabled`) and pass it through `findNewVersion`/`getTagCandidates`.
- 🐛 Fixed issue `#498` by restoring correct “current vs newest” visibility for pinned prerelease tags without setting `updateAvailable`.
- 🐛 Added/expanded regression tests across model validation, trigger dispatch gating, tag-candidate logic, version finding, and UI badge rendering.

## Concerns

- Confirm API/UI backwards compatibility for the new optional `result.updateInsight` field.
- Ensure prerelease→GA insight widening remains strictly informational (never leaks into actionable candidate paths).
- Validate sorting/matching correctness for additional registry tag formats and edge-case suffix precision/shape parsing not covered by current tests.
- Verify memoization/regression coverage fully spans all view modes and the `v-memo` equality behavior around insight changes.


scttbnsn:
@coderabbitai Both addressed in b04ea133 + 9c8ad8e8: the prerelease→GA exception is now scoped to the insight path via an allowPrereleaseToGA flag that only computePinGateInsight sets (added a red-first test proving loose no longer surfaces bare 1.5.2 as an actionable candidate for a 1.5.2-rc.1 pin), and getContainerViewMemo now includes updateInsight tag+kind with a memo-diff test.

biggest-littlest (approved):
Reviewed the informational-channel design and the trigger-safety regression test. LGTM.

ALARGECOMPANY (approved):
Verified the pin-gate scoping and the loose-mode suffix fixes. LGTM.

PR #520, opened 2026-07-13
🐛 fix(audit): suppress security scan lifecycle churn

scttbnsn:
## Summary

- keep security scan persistence and container SSE refreshes intact
- emit generic container-update audit rows only for meaningful lifecycle, image, update, health, error, or policy transitions
- scope transition state by persisted compose-aware identity with safe legacy fallback
- prune orphaned state after 30 days and clear exact state on removal
- document the v1.6 behavior in the changelog

## Root cause

Scheduled security scans update security.scan.scannedAt. The store correctly emits dd:container-updated so the UI refreshes, but the audit subscriber treated every event as a lifecycle transition. A 50-container scan therefore wrote 50 unchanged container-update audit rows per cycle and broke the #321 store-size acceptance bound.

## Verification

- focused audit suite: 35 tests
- audit/index integration: 17 tests
- store/container integration: 256 tests
- repository pre-push gate: green
  - backend and UI coverage: 100%
  - backend and UI builds: green
  - 88 script tests, 31 workflow tests, 37 website script tests
  - Biome, Qlty gate, UI typecheck, and Zizmor: green
- independent two-pass implementation review: clean after compose-sibling and fleet-cap regressions were added

## Acceptance impact

This addresses the audit-growth failure found in the superseded #321 operational run. The final 24-hour NAS acceptance remains intentionally last, after all v1.6 code is frozen.

✨ Added
- Lifecycle-aware container audit deduplication cache using deterministic “lifecycle audit signatures”.
- Compose/agent/watch-er scoped identity for dedupe, with safe legacy fallback.
- Cache TTL pruning (30 days) and lifecycle cache cleanup on container removal.
- v1.6 changelog documentation.
- Expanded audit subscription test harness to fully register lifecycle handlers (added/updated/removed) and expose lifecycle handler payload helpers (including `makeLifecyclePayload`).

🔧 Changed
- `container-update` audit insertion is now conditional: recorded only when lifecycle/image/update/health/error/update-policy/policy-related transitions actually change, suppressing “security-only” scheduled scan refresh churn.
- Lifecycle state is seeded only after successful audit persistence for add/changed-update paths (prevents lifecycle state from being seeded on failed insertions).
- `container-updated` details population was adjusted to avoid unnecessary/incorrect prior unconditional fields; status-derived transitions are tracked via signature comparison instead.
- Lifecycle cache entries are pruned/cleared deterministically:
  - stale entries removed by TTL
  - orphaned lifecycle state cleared after 30 days
  - lifecycle state cleared when containers are removed
- Expanded test coverage for:
  - suppression after lifecycle baseline (including partial baselines)
  - retry behavior on failed audit writes (failed insert can’t seed/suppress lifecycle state)
  - stable status transitions recorded while repeated security refreshes are suppressed
  - normalize update-policy list ordering before comparing lifecycle state
  - identity scoping (including sibling behavior) and legacy fallback safety
  - “fails open” behavior when identity cannot be derived
  - avoiding churn for very large fleets (>10k)

🐛 Fixed
- Prevented unchanged container-update audit rows caused by scheduled security scans, while preserving security scan persistence and container SSE refresh behavior.
- Ensured persistence-ordering correctness: lifecycle cache advances only after audit persistence succeeds (add/changed-update paths).
- Failure-injection tests confirm lifecycle state can’t be seeded/suppressed by failed inserts.

🗑️ Removed
- Unconditional generic `container-update` audit insertion behavior that caused audit churn.

⚠️ Breaking
- None indicated.

🔒 Security
- Keeps security scan persistence/SBOM-related state advancing without generating noisy generic audit rows during scheduled refreshes.

- Confirm lifecycle cache identity derivation is consistent at all production call sites (and that the legacy fallback cannot merge unrelated containers).
- Verify persistence-ordering fix remains correct under retries/failure injection (no lifecycle advancement on failed audit writes).
- Validate TTL pruning and container-removal cleanup don’t accidentally evict still-relevant lifecycle state for sibling/compose scenarios.
- Ensure dedupe behavior “fails open” only when identity truly can’t be derived, without reintroducing churn.


scttbnsn:
Addressed both findings in a11e763b:

- Lifecycle cache state now advances only after audit persistence succeeds on both add and changed-update paths. Two failure-injection regressions prove a failed insert is retried and cannot seed/suppress state.
- Added the shared makeLifecyclePayload helper and replaced the repeated lifecycle test fixtures/factories.

Validation: 312 focused integration tests, backend build, enforced pre-push gate, 100% app/UI coverage, and app/UI builds all pass. The isolated DinD operational smoke also held audits at 7 -> 7 and container-update at 0 across two on-demand refreshes and two scheduled cycles while scan/SBOM state advanced; no host socket was exposed and all smoke resources were removed.

@coderabbitai review

PR #532, opened 2026-07-14
🔧 chore(ci): allow manual final-matrix dispatch

scttbnsn:
## Summary

- add manual dispatch to CI Verify and Playwright on the default branch
- enable CodeQL, fuzz, web, DAST, Cucumber, load, and browser gates for an explicitly selected frozen ref
- preserve path-filter diagnostics while overriding their skip decisions for manual release-candidate validation
- keep dependency review on the exact-head PR check

## Why

GitHub only permits manual workflow dispatch when the workflow advertises that trigger on the default branch. This focused bootstrap is therefore required before the complete v1.6 matrix can run against the frozen release ref.

After this lands, the sanitized v1.6 release PR will be rebuilt from the new immutable main SHA plus the accepted dev/v1.6 tree.

## Verification

- TDD red confirmed both workflows lacked manual dispatch
- 33 workflow tests pass
- Biome passes with the exact main lockfile
- Actionlint passes
- Zizmor reports zero findings
- no runtime, dependency, release, or planning files changed

Companion v1.6 behavior PR: #531.

### Changelog

- ✨ Added `workflow_dispatch` support to CI Verify and Playwright workflows.
- 🔧 Updated `dorny/paths-filter` “Filter paths” base handling for manual runs (defaulting to `main`) and broadened key job `if` conditions to include `workflow_dispatch` (including release-candidate validation paths).
- 🔧 Updated CI Verify + Playwright e2e workflow tests and workflow-YAML typings to assert manual dispatch bypasses path gating where intended (including `dependency-review` `if` conditions) and to validate combined load-test dispatch logic.
- 🔒 Kept dependency review limited to exact-head PR checks.
- 🔧 Verified the workflow/test matrix with 33 passing workflow tests plus Biome, Actionlint, and Zizmor.

### Concerns

- Confirm the `workflow_dispatch` path-filter comparison base (`main`) matches the intended frozen/ref selection behavior for manual release-candidate validation.
- Confirm every manually dispatched gated job (and their internal steps like Playwright browser cache/download) follows the expected dispatch vs. path-change semantics.
- Confirm `dependency-review` continues to use exact-head logic for the relevant event types when run via manual dispatch.

### DIFF
--- a/.github/tests/ci-verify-workflow.test.ts
+++ b/.github/tests/ci-verify-workflow.test.ts
@@ -27,6 +27,7 @@ interface WorkflowJobStep {
 interface WorkflowDefinition {
   env?: Record<string, string>;
   jobs?: Record<string, WorkflowJob>;
+  on?: Record<string, unknown>;
 }
 
 interface LefthookCommand {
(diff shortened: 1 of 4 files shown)


scttbnsn:
Local gate follow-up on exact head 8999dcb1:

- the initial aggregate run exposed borrowed v1.6 node_modules in the isolated v1.5.2 worktree, not a source failure
- replaced the borrowed root, app, and UI dependencies with exact lockfile installs
- app and UI coverage both meet 100%
- app and UI builds pass
- 33 workflow tests pass
- Biome, Qlty baseline, Actionlint, and Zizmor pass

Hosted exact-head CI remains authoritative and is running now.

scttbnsn:
CodeRabbit nitpick verified and intentionally not adopted: these helpers are file-local because each test binds a different workflow path and uses a different job/step shape. Extracting a new shared test-utility module would expand this release-bootstrap change into unrelated test-architecture refactoring, while the existing duplication predates this PR. The companion v1.6 PR #531 received no actionable review comments for the same scoped pattern.

biggest-littlest (approved):

ALARGECOMPANY (approved):

Issue #565, opened 2026-07-20
Checking for updates again resets the maturity time


ciltocruz:
### Drydock version

1.6.0-rc.2

### What happened?

When I click the "check for updates" button in a container, the maturity settings are deleted.

Before:

<img width="721" height="541" alt="Image" src="https://github.com/user-attachments/assets/6b5b233b-d4e4-45ed-984a-ac6174c1371d" />

After:

<img width="864" height="538" alt="Image" src="https://github.com/user-attachments/assets/34fad979-ff75-44f4-b248-10c08267394f" />

### What did you expect?

Don't reset the config

### Steps to reproduce

1. Set madurity for 1 container.
2. Recheck updates
3. Madurity reset ?

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

scttbnsn:
Thanks for the clear repro, dug into this and found the mechanism. Good news first: your maturity settings aren't actually deleted. The mode and minimum-days config survive the recheck untouched. What resets is the countdown clock, which reads the same as losing the setting.

Here's what happens: the maturity gate counts from when drydock first detected the update candidate. A manual recheck recomputes the candidate from the registry, and the "did the candidate change" comparison currently looks at more than the candidate's identity. It also compares display metadata like the suggested tag and the image's created timestamp. If any of those come back slightly different, drydock treats it as a brand-new candidate and restarts the clock from now, even though the actual update (same tag, same digest) never changed.

That's a bug. Restarting the soak when a genuinely new candidate shows up is intentional; restarting it because cosmetic metadata wobbled between scans is not. The fix is to narrow that restart check to the candidate's real identity (tag + digest), and it's queued for the next release candidate. This isn't something rc.2 introduced by the way, it's been in the maturity gate since 1.5.x.

One question to confirm the trigger on your side: which registry are those containers from (Docker Hub, GHCR, lscr, something else)? For Docker Hub and GHCR the published date usually floors the clock and masks this, so knowing your provider helps me confirm the repro matches.

scttbnsn:
Fix is up in #568, queued for the next release candidate. The registry question above still stands when you get a chance.

ciltocruz:
> One question to confirm the trigger on your side: which registry are those containers from (Docker Hub, GHCR, lscr, something else)? For Docker Hub and GHCR the published date usually floors the clock and masks this, so knowing your provider helps me confirm the repro matches.


The test was performed on the Drydock image:
[...]

PR #568, opened 2026-07-20
🐛 fix(store): only restart the maturity clock when the update candidate identity changes

scttbnsn:
## What

Fixes the #565 report: a manual per-container "check for updates" restarted the maturity soak countdown, which reads to the user as their maturity settings being wiped (the config itself was never touched — only the clock).

Root cause: `getUpdateLifecycleTimestamp`'s restart branch gated on `Container#resultChanged`, which compares `tag`, `suggestedTag`, `digest`, AND `created`. The last two are display metadata that can drift between scans (a manual recheck bypasses the registry poll cache), so a cosmetic wobble re-stamped `updateDetectedAt` to now and the countdown snapped back to the full window. Registries with trusted publish dates (Docker Hub/GHCR) mask this via the publishedAt floor; everyone else is fully exposed.

Fix: new `hasCandidateIdentityChanged()` comparing only `tag` + `digest`; the lifecycle restart branch uses it directly. `resultChanged`/`hasResultChanged` are untouched — the trigger-notification `changed` flag in `container-processing.ts` must stay sensitive to `suggestedTag`/`created` changes for notification content. Intentional behavior is preserved: a genuinely new candidate (tag or digest change) still restarts the soak.

## Why

Diagnosed against the live module (drove `updateContainer` through the exact recheck sequence): identical results leave the clock alone; a `suggestedTag`-only or `created`-only difference reset it every time. Pre-existing since the v1.5.x maturity gate (restart branch from v1.5.1-rc.1, comparison set from v1.5.0) — not an rc.2 regression. Targets the v1.6.1/rc.3 train.

## Tests

Unit coverage for `hasCandidateIdentityChanged` (undefined handling both directions, each field diff, metadata-only diff → false) plus three store-level regression tests on `updateContainer`: metadata-only recheck preserves `updateDetectedAt`; tag change resets; digest change resets. Full pre-push gate green; `app/model/container.ts` and `app/store/container.ts` at 100% on all four metrics.

## Changelog

- 🐛 Fixed maturity soak countdowns restarting when only display metadata changes.
- ✨ Added `hasCandidateIdentityChanged()` to compare update candidate identity using only `tag` + `digest`.
- 🔧 Updated lifecycle timestamp logic to restart only on candidate identity (`tag`/`digest`) changes (leaving existing `resultChanged` behavior for trigger notifications intact).
- ✨ Added unit + store regression tests covering metadata-only changes, tag/digest changes, and undefined handling.
- 🔒 Security: Updated `brace-expansion` minimum safe version (5.0.7) and related e2e override validation.
- 🔧 Dependency: Bumped pinned `js-yaml` override (4.2.0 → 4.3.0) and `brace-expansion` override (5.0.6 → 5.0.7).

- Ensure maturity restart logic uses the correct “raw update” `result` shape consistently across persistence/rechecks.
- Verify `hasCandidateIdentityChanged()` undefined/missing-field semantics align with actual runtime inputs.
- Confirm trigger notification paths still rely on existing `resultChanged` behavior as intended.
- Run/verify unit + store tests and e2e security suite with the updated dependency overrides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Manual recheck no longer restarts the maturity countdown on display-only metadata drift** ([#565](https://github.com/CodesWhat/drydock/issues/565)). A per-container recheck bypasses the registry poll cache, so the suggested tag and image created date can wobble between scans even when the update candidate itself hasn't changed, falsely resetting the soak. The lifecycle clock now restarts only when the candidate's actual identity — tag or digest — changes; suggested tag and created date no longer factor into the restart decision.
+
 ### Security
 
 - **Anonymous access now fails closed on upgrades, not just fresh installs.** An instance with no authentication configured — or with anonymous auth enabled but unconfirmed — refuses to start instead of logging a warning and serving an open dashboard. If you run an intentionally open instance, set `DD_ANONYMOUS_AUTH_CONFIRM=true`; otherwise configure `DD_AUTH_BASIC_<name>_USER`/`_HASH` before upgrading.
--- a/app/model/container.test.ts
+++ b/app/model/container.test.ts
@@ -520,6 +520,55 @@ test('testable_resultChangedFunction should compare created timestamps and handl
   ).toBe(true);
 });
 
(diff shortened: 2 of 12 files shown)


scttbnsn:
On the mock-setup nitpick: leaving as-is. The sibling tests in this file inline their fixture setups the same way, and a helper shared by only the three new tests would diverge from the file idiom for marginal savings.

PR #572, opened 2026-07-21
🚀 release: v1.6.0-rc.3

scttbnsn:
Promotes the v1.6.0-rc.3 content from `dev/v1.6` to `main` for the release cut.

rc.3 delta over rc.2:
- **#566 display-honesty batch** — update-status vocabulary ("Digest update", Unknown badge, "Security hold"), pin glyph driven by the real pin-gate verdict (`tagPinGated`), insight-only pinned rows read Current everywhere (table + detail panel), one-clock maturity panel, column-picker auto-hide annotations, registries deep-link fix (#556), banner pluralization, tooltip freshness.
- **#567 security hardening** — anonymous auth fails closed on upgrades, HTTP trigger SSRF guards, full WebSocket origin validation, `drydock.sid` session cookie, owner-only store perms, pinned icon CDN revisions, CI secret scans.
- **#568 maturity-clock restart fix** (#565) — the soak clock restarts only when the update candidate identity changes.
- **#569 docs freshness** — auth/SSRF/WS/UI copy corrected to shipped v1.6 behavior.
- **#571 release identity bump** — dated CHANGELOG heading, rc.3 across all public version surfaces, guard tests in lockstep.

After merge: dispatch `release-cut.yml` from `main` with `release_tag=v1.6.0-rc.3`.

PR #573, opened 2026-07-21
✅ test(e2e): align pinned-display and maturity-clock expectations with #566

scttbnsn:
The 🎭 E2E: Playwright job on release PR #572 caught two specs still asserting pre-#566 display behavior. Both are intentional changes from the display-honesty batch, not regressions:

- **#406 maturity countdown**: the one-clock panel collapsed the duplicate countdown into a single `{countdown} · unlocks {date}` line, so the `Lifts at` regexes never match. Updated to `unlocks`.
- **#498 pinned card state**: pinned left the Update column — insight-only cards read **Current**, and the insight-kind badge was retired in favor of the persistent pin glyph on the Tag cell. The spec now asserts `Current` plus the `container-tag-pinned-glyph` in both the table row and the card, and the fixture sets `tagPinGated` (the backend gate verdict the glyph keys on).

Verified locally against the QA env (drydock:dev, :3333): all 4 tests in the spec pass.

Unblocks #572.

## Changelog

- 🔧 Changed maturity countdown assertions to expect `unlocks` wording with updated minute intervals.
- 🔧 Changed pinned-card assertions to expect `Current` and the persistent pin glyph.
- ✨ Added `tagPinGated = true` to the pinned-container fixture.
- 🗑️ Removed the outdated update-insight-kind badge assertion.

### DIFF
--- a/e2e/playwright/v16-modes-pins.spec.ts
+++ b/e2e/playwright/v16-modes-pins.spec.ts
@@ -158,11 +158,11 @@ test.describe('v1.6 update modes, scheduling, and pinned tags', () => {
     await expect(condition).toContainText(
       'The candidate must remain unchanged until the stabilization period ends.',
     );
-    await expect(condition).toContainText(/6m\s*·\s*Lifts at /);
+    await expect(condition).toContainText(/6m\s*·\s*unlocks /);
     await expect(panel.locator('[data-test="update-status-manual-cta"]')).toBeEnabled();
 
     await page.clock.fastForward(60_000);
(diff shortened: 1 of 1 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
Superseded by #574, which includes this PR’s exact E2E fix commit (`460fb91a`) plus the complete v1.6 review remediation and release hardening. Keeping the full integration on one reviewed path avoids splitting the candidate tree.

PR #594, opened 2026-07-24
🐛 fix(deps): stop Renovate shipping package.json bumps without the lockfile

scttbnsn:
## What was happening

Renovate bumped `ui/package.json` and `apps/web/package.json` and never wrote either lockfile, so `npm ci` rejected both branches:

```
npm error code EUSAGE
`npm ci` can only install packages when your package.json and package-lock.json
or npm-shrinkwrap.json are in sync.
```

Every workflow that installs those workspaces died at its own install step — Test & Coverage, Playwright, Cucumber, and the site build — so **one bad lockfile read as four unrelated CI failures**, each minutes into a job, with an error that never names Renovate.

A census of the last 200 runs: 46 failures, **40 of them on `renovate/*` branches**, and 24 of the 46 were this single bug. `app/`, `e2e/` and `apps/demo/` were updated correctly in the same sweep, which is why it read as flake.

```
renovate/npm-minor-patch-ui         ui/package.json                           ← no lockfile
renovate/npm-minor-patch-appsweb    apps/web/package.json                     ← no lockfile
renovate/npm-minor-patch-e2e        e2e/package.json  e2e/package-lock.json   ← fine
renovate/npm-minor-patch-app        app/package.json  app/package-lock.json   ← fine
```

## `skipInstalls: false`

`skipInstalls` is Renovate's fast path: patch the lockfile directly and skip the install when it believes it can. That is what silently no-opped. Turning it off costs a real npm install per update and is correct every time.

**This diagnosis is inferred.** Renovate reported no artifact-update problem in either PR body or on the dependency dashboard, and its run logs are behind the Mend portal. Config validated against the live `renovate-schema.json` — `skipInstalls` accepts `boolean|null`, no unknown keys.

## The guard, which does not depend on that being right

`scripts/lockfile-sync.test.mjs` reimplements the comparison `npm ci` makes — lockfile `packages[""]` against `package.json` — across every committed lockfile. It runs in the pre-push hook and as the first CI step after the root install, **before any workspace install**, and names the drifted packages:

```
npm ci will fail with EUSAGE in these workspaces:
  ui
    dependencies.vue: package.json=3.5.40 lockfile=3.5.39
Regenerate each one with `npm install --package-lock-only` in that directory and commit the result.
```

Roots come from `git ls-files -- '*package-lock.json'`, so a seventh workspace is covered the day its lockfile lands — no list to keep current. Verified in both directions, and all six current roots are checked.

## Already unblocked

I regenerated both lockfiles and pushed them to the bot branches. #485 and #510 are green for the first time in days — `npm ci` in `apps/web` now succeeds and pulls `lefthook@2.1.10`, the bumped version. Those hand-fixes would be overwritten on Renovate's next run; this PR is what makes it stop happening.

## Changelog

- ✨ Added `scripts/lockfile-sync.test.mjs` to detect dependency drift between tracked `package.json` and `package-lock.json` files.
- 🔧 Changed Renovate to disable `skipInstalls`, ensuring dependency updates regenerate lockfiles.
- 🔧 Changed pre-push and CI checks to run lockfile consistency validation.
- 🔧 Regenerated the `ui` and `apps/web` lockfiles.

## Concerns

- Verify all contributors have Node.js and Git available when running the new check.
- Confirm lockfile discovery and dependency comparisons support every workspace layout.


scttbnsn:
@coderabbitai review

PR #596, opened 2026-07-24
🔧 chore(sync): advance main to dev/v1.6

scttbnsn:
Routine dev → main sync, per the branch model (main only advances by merging from dev/v1.6). This one carries the CI-hardening and bot-config batch:

- #590/#592 drift guard + integration-branch targeting (already on dev, now visible to the bot: Renovate reads renovate.json from the default branch, so baseBranchPatterns and skipInstalls only take effect on main)
- #594 skipInstalls: false + the lockfile-sync guard (runs npm's own ci --dry-run check per workspace)
- #593 .planning untracked + tracked-ignored-files guard
- #545 Crowdin translation sync with all 21 review comments fixed upstream
- Five Renovate dependency batches (#484/#485/#487/#489/#510)
- postcss override 8.5.22 (GHSA-r28c-9q8g-f849)

After merge, `git diff origin/main origin/dev/v1.6` must be empty (tree equality — the invariant release-cut enforces).

PR #599, opened 2026-07-26
🐛 fix(ci): make pipefail dead-fallback handlers actually reachable

scttbnsn:
The Crowdin sync failed on main on 2026-07-24 with a bare `exit 1` and no message. Root cause turned out to be a bug class rather than a one-off, so this fixes both instances of it and adds coverage.

## The pattern

Inside a `set -e`/`pipefail` region, a command that exits non-zero on an *empty* result aborts the step before an explicitly written fallback can run. The fallback is dead code. The tell: someone wrote handling for the empty case, and the empty case can't reach it.

## 1. Crowdin base-branch resolution

`i18n-crowdin.yml` resolves the highest `dev/vX.Y` on origin and falls back to the default branch when there is none, so translations always have somewhere to land between a GA and the next dev branch being cut. Under `set -euo pipefail` a no-match `grep` exits 1, which fails the pipeline, which fails the assignment, which aborts the step. The `if [ -z ]` fallback below it could never run.

It fired for real: merging sync PR #596 briefly deleted `dev/v1.6` via auto-delete-head-branches, the push-to-main sync run found no `dev/v*`, and the job died.

Guards the grep only, so a failing `ls-remote` (network, auth) still propagates and fails loudly instead of silently retargeting translations at main. That distinction matters more than the fix itself.

## 2. Load-test report resolution

Sweeping the other workflows for the same pattern turned up 8 more sites in `ci-verify.yml`:

```
report="$(find artifacts/load-test/<mode> ... | sort -rn | head -n1 | cut -d' ' -f2-)"
```

`find` exits 0 on a directory that exists but is empty — the case the downstream scripts handle. It exits 1 when the directory is *missing*, and GitHub runs these steps under the default `bash -eo pipefail`, so that aborts the step. `summarize-load-test-report.sh` and `check-load-test-correctness.sh` both open with an `if [ -z "${REPORT}" ]` block written for exactly this, and neither gets to run.

Reachable, not theoretical: these steps are `if: always()`, and `run-load-test.sh` only does `mkdir -p` on the artifact dir at line 168 — after the docker build, compose up, and the health-check wait that exits 1 at line 136. A load test that fails the way load tests usually fail leaves the directory absent, and the step meant to explain that dies with a bare exit 1.

## Coverage

`crowdin-base-branch-resolver.test.ts` executes the real `id: base` run block extracted from the workflow against a stubbed git, rather than asserting on workflow text, so it tracks the actual logic instead of a copy that can drift. Three cases: no dev branches falls back and exits 0; several picks `dev/v1.10` over `v1.9` (lexical sort gets this wrong); failing `ls-remote` still exits non-zero. That last one guards against "fixing" the bug by swallowing all errors, which would be worse than the bug.

Verified it fails against the pre-fix snippet with `expected 1 to be +0`, and that the other two still pass — the bug only affects the one path.

`ci-verify-find-pipefail-guard.test.ts` discovers the find sites by walking the parsed workflow rather than pinning line numbers, so it also catches a new unguarded `find` added later. Carries a sanity assertion so it can't pass vacuously if the shape stops matching. Verified it catches a revert by naming the exact offending step.

Executing the load-test steps for real would need docker plus Artillery artifacts, so that one is a static assertion by design — noted in a comment in the file.

## Notes

- Full pre-push gate green locally, including the 100% coverage gate and zizmor.
- Related but separate: `main` is currently failing its scheduled CI Verify on `brace-expansion@5.0.7` (CVE-2026-14257). That is already fixed on `dev/v1.6` at 5.0.8 — it just needs the next dev → main sync, no code change.
- Also closed #598 and deleted `l10n_crowdin`; it was cut before #545 landed and had gone conflicting. It regenerates from the dev tip on the next push here.

## Changelog

- 🐛 Fixed Crowdin base-branch resolver so “no matching dev/vX.Y” correctly falls back to `DEFAULT_BRANCH` under `set -euo pipefail`, while `git ls-remote` failures still propagate.
- 🐛 Fixed CI load-test report discovery by guarding each `find artifacts/load-test/<site>/...*.json ...` command-substitution pipeline so missing artifact directories don’t abort later empty-report handling under `pipefail`.
- 🔧 Changed `test/cpu-bench.sh` to guard the per-container `grep "^$c "` used to compute stats so missing samples don’t stop the script under `set -euo pipefail`.
- ✨ Added Jest coverage for Crowdin fallback behavior, highest `dev/vX.Y` selection, and `ls-remote` error propagation.
- ✨ Added static workflow analysis coverage to ensure every load-test report lookup assignment site uses a fully pipefail-safe `find` guard (including later pipeline stages).

## Concerns

- Ensure the workflow-scan regex continues to correctly identify the intended `find`-into-variable command-substitution assignments and doesn’t miss new sites.
- Verify the static “expected eight-site count” remains in sync with the number of load-test report lookup assignments across CI (ci/behavior/stress, summarize/correctness/regression).

### DIFF
--- a/.github/tests/ci-verify-find-pipefail-guard.test.ts
+++ b/.github/tests/ci-verify-find-pipefail-guard.test.ts
@@ -48,6 +48,31 @@ interface FindAssignmentSite {
   line: string;
 }
 
+// Reconstructs logical shell lines from a `run:` block by folding backslash
+// continuations, so a site reformatted across `\` line breaks is still
+// discovered by the single-line FIND_ASSIGNMENT_LINE regex below. A line
+// ending in a trailing `\` (optionally followed by trailing whitespace)
+// continues onto the next physical line.
+function joinContinuations(run: string): string[] {
+  const logicalLines: string[] = [];
+  let pending = '';
+
+  for (const rawLine of run.split('\n')) {
+    const continued = /\\\s*$/.test(rawLine);
+    const chunk = continued ? rawLine.replace(/\\\s*$/, '') : rawLine;
+    pending += (pending ? ' ' : '') + chunk.trim();
+
+    if (!continued) {
+      logicalLines.push(pending);
+      pending = '';
+    }
+  }
+
+  if (pending) logicalLines.push(pending);
+
+  return logicalLines;
+}
+
 function findFindAssignmentSites(): FindAssignmentSite[] {
   const workflow = loadWorkflow();
   const sites: FindAssignmentSite[] = [];
@@ -56,7 +81,7 @@ function findFindAssignmentSites(): FindAssignmentSite[] {
     for (const step of job.steps ?? []) {
       if (!step.run) continue;
 
-      for (const line of step.run.split('\n')) {
+      for (const line of joinContinuations(step.run)) {
         const match = line.match(FIND_ASSIGNMENT_LINE);
         if (!match) continue;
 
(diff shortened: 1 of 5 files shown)


scttbnsn:
Good catch on the anchored regex, that was a real hole. Fixed in 3d651b13.

Went with requiring exactly one `find` per assignment rather than parsing each stage. Every current site has one, and a second should fail the test so someone actually looks at it. Verified the `{ find a || true; } | find b ...` case passes before the change and fails after, so it's covering the thing you flagged rather than just looking stricter.

PR #602, opened 2026-07-26
📝 docs(release): prepare v1.6.0-rc.6

scttbnsn:
Prepares the release surfaces for cutting `v1.6.0-rc.6`. No runtime code changes.

## Why now

rc.5 was cut on 2026-07-23. Since then `dev/v1.6` picked up **#587, the maturity-cleared notification** — a new feature, not a fix, with a new `maturity-policy` model, ~170 lines in `Trigger.ts`, ~100 in `store/container.ts`, and a new background sweep on a new `DD_MATURITY_SWEEP_CRON` default. On top of that the delta carries a full dependency bump wave across all six lockfiles plus a Crowdin translation resync.

That's too much unsoaked change to take straight to GA, so it gets an RC first.

## What's here

**Changelog.** Promotes `[Unreleased]` to `[1.6.0-rc.6] — 2026-07-26` so `release-cut.yml`'s CHANGELOG validation resolves an entry for the tag, and backfills the rest of the delta since rc.5: the dependency wave, the translation resync, and the three CI/repo fixes (#599, #594, #593) that landed without changelog entries of their own.

**Release identity.** The usual 16-file bump: version badge, `site-config.ts`, the roadmap entry in `site-content.ts`, demo mock fixtures, the API doc response samples, the quickstart tag matrix, a new highlights section on the updates page, and the three pinned identity tests under `scripts/`.

All eight version files that `release-cut.yml` checks already sit at the `1.6.0` base, so no `package.json` bump is needed for this RC.

## Verification

- `node --test scripts/*.test.mjs` — 133 pass, 0 fail
- Full pre-push gate green end to end: qlty, qlty-smells, scripts-test, workflow-tests, typecheck-ui, web-scripts-test, coverage (230s), build, zizmor
- `main` and `dev/v1.6` trees were in sync before this branch, so the pre-cut drift guard only needs the one sync after merge

### ✨ Added
- Added `v1.6.0-rc.6` highlights (maturity-cleared notifications, dependency + translation resync, CI scripting robustness) for July 26, 2026.

### 🔧 Changed
- Promoted the `v1.6.0-rc.6` release surface by updating version identity references from `v1.6.0-rc.5` → `v1.6.0-rc.6` across README, site config/content, roadmap/demo content, API documentation samples, quickstart tags, and the updates page.
- Updated mock/demo data to report `1.6.0-rc.6` consistently (agent, audit, container, server, and API handler fixtures).
[...]

PR #603, opened 2026-07-26
🔧 chore(sync): advance main to dev/v1.6 for the v1.6.0-rc.6 cut

scttbnsn:
Routine dev → main sync ahead of cutting `v1.6.0-rc.6`. Per the branch model, `main` only advances by merging from `dev/v1.6`, never by taking independent commits, and the cut is dispatched from `main`.

Opened from a throwaway `sync/main-rc6` branch rather than from `dev/v1.6` directly. Auto-delete-head-branches deleted `dev/v1.6` when sync PR #596 used it as the head, which is what broke the Crowdin sync.

## Content

The entire delta is the rc.6 prep merged in #602: the dated `[1.6.0-rc.6]` changelog entry and the 16-file release identity bump.

## Conflict resolution

The three-way merge conflicted on `CHANGELOG.md`. That's an artifact of squash-merge-only ancestry — `main` picked up the same content through an earlier squash, so the merge reads both sides as independently adding it.

Resolved with `git merge -s ours origin/main`, which takes `dev/v1.6`'s tree wholesale. Verified that loses nothing:

- No file exists only on `main`
- The complete content diff between `main` and `dev/v1.6` is exactly the 17 rc.6 prep files, nothing else

After the merge, `git diff HEAD origin/dev/v1.6` is empty, so this branch's tree is byte-identical to `dev/v1.6`. That's the invariant the release-cut drift guard asserts.

## Verification

- Full pre-push gate green: qlty, qlty-smells, scripts-test, workflow-tests, typecheck-ui, web-scripts-test, coverage, build, zizmor
- `renovate.json` already targets `dev/v1.6`, so the drift guard's bot-target assertion passes
- All eight version files sit at the `1.6.0` base that the tag validation expects

Next after merge: dispatch `release-cut.yml --ref main -f release_tag=v1.6.0-rc.6`.


biggest-littlest (approved):
Sync only. Tree verified identical to dev/v1.6.

ALARGECOMPANY (approved):
Approved. Content matches dev/v1.6 exactly.

Issue #606, opened 2026-07-26
Manifest "created" fetch failures are swallowed, making result.created nondeterministic


scttbnsn:
Found by a sweep for the identity-drift defect class behind #565. This one isn't a bug on its own so much as the condition that makes several others fire.

## The defect

`app/registries/Registry.ts:407` (`fetchImageCreatedFromManifestConfig`) and `:437` (`fetchImageCreatedFromBlob`) each catch any failure and return `undefined`. These run on the schemaVersion-2 single-manifest path, which is most providers.

So a transient registry 5xx, a 429 storm, or a network blip that outlasts the retry budget silently produces a result with no `created`, even though `tag` and `digest` are byte-identical to the previous poll. Nothing is logged above debug.

## Why it matters

`result.created` is not display-only. It participates in several "is this the same update candidate?" comparisons, and when it wobbles those comparisons produce spurious "changed" verdicts:

- **Notification dedup.** `computeResultHash` (`app/store/notification-history.ts:54`) hashes `created` unconditionally, so a wobble breaks `once: true` and fires a duplicate notification for an update the user already saw.
- **Recreate lifecycle stash.** `getResultSignature` (`app/store/container.ts:117`) compares `created`, so a wobble across a container recreate discards the stash and silently restarts the maturity soak.

The primary clock is already protected on the common path by `hasCandidateIdentityChanged` (`app/model/container.ts:926`), which correctly ignores `created` whenever a digest is present. The two consumers above never got that treatment.

## Relationship to the other fixes

Both consuming bugs are being fixed directly, by consolidating all three comparators onto one shared candidate-identity helper. That removes the harm.

This issue is for the underlying cause: a failed metadata fetch should be distinguishable from an image genuinely having no `created` date. Right now those are the same `undefined`, and the failure is invisible. Fixing the consumers makes the current symptoms go away but leaves that ambiguity in place for the next consumer that reaches for the field.

Not a regression, and no user-visible symptom once the consumers land, so this is v1.6.1 or later.

Issue #605, opened 2026-07-26
Eligibility can show "Agent Mismatch" during the agent trigger registration window


scttbnsn:
Split out of [discussion #325](https://github.com/CodesWhat/drydock/discussions/325), where a reporter asked about this and never got a correct answer.

## What they saw

Same container (Prowlarr, on a remote agent host) showed `AGENT MISMATCH` in one screenshot and `TRIGGER FILTERED` in another taken after upgrading rc.14 to rc.15. Nothing about the container, host, registry, or image changed between them.

They were originally told this was two different containers landing in different branches. That was wrong. Both screenshots show the same row.

## What's actually happening

The two states are mutually exclusive by design:

- `agent-mismatch` fires when no docker/dockercompose trigger is reachable for that container's agent at all
- `trigger-not-included` fires once a reachable trigger *is* found but its include label doesn't name the container

That branch logic is byte-identical between rc.14 and rc.15 (`git diff v1.5.0-rc.14 v1.5.0-rc.15 -- app/model/update-eligibility.ts` is a `makeBlocker()` wrapper and a severity map, nothing else). `app/api/docker-trigger.ts`, which holds the function whose verdict actually flips the pill, doesn't appear in that diff at all. So no code change caused it.

The cause is that eligibility is recomputed live on every read and never cached. `computeUpdateEligibility` is called fresh at `app/api/container/handlers/list.ts:253`, `app/api/sse-container-enrichment.ts:50`, and `app/updates/request-update.ts:295`, each pulling `context.getTriggers()` at call time.

So the displayed state reflects whatever the controller believes is reachable for that agent at that instant. During the window after a restart or reconnect, before the agent's docker trigger has finished registering, `agent-mismatch` is what a user sees. Once registration completes it flips.

## Why this is worth fixing

The condition is transient and self-correcting, but it's indistinguishable from a real misconfiguration. `agent-mismatch` is a `hard` blocker, so it also disables the Update button for the duration.

`git log --all --grep` finds nothing that has ever diagnosed this, so it's undiagnosed rather than known-and-accepted.

## Still present on dev/v1.6

The pill was replaced by the Update Status panel (#566), but the underlying condition is unchanged and still computed live. `ui/src/composables/useUpdateStatus.ts:127-133,177` maps `agent-mismatch` to a condition with a docs deep link, so the same window surfaces the same way.

## Not reproducible from logs

[...]

Issue #604, opened 2026-07-26
Maturity clock: swallowed GHCR/LSCR auth errors and divergent age thresholds


scttbnsn:
Split out of #556. The symptom reported there was resolved (display-only, confirmed by the reporter), but the investigation surfaced separate latent defects that were never tracked on their own. Filing so they don't get buried when #556 closes.

None of these are regressions. They predate v1.6.

## 1. GHCR/LSCR auth failures are swallowed at debug level

`app/watchers/providers/docker/image-comparison.ts:341-343` and `:425-429` catch every failure from `getImagePublishedAt` and log at `debug` only. That includes a 401/403 from GitHub's Packages API when no `read:packages` token is configured, or when a fine-grained token lacks the scope.

The maturity gate then silently falls back from the trusted registry push date to `updateDetectedAt`. The clock keeps running, so nothing looks broken, but it's measuring the wrong thing: how long *we've known about* the update rather than how long the image has actually existed. That defeats the supply-chain-timing guarantee the gate is there to provide.

There's no user-visible signal at all. Diagnosing it means enabling debug logging and grepping. This was the actual root cause chased through most of #556.

## 2. Two of four update-age clocks still ignore per-container policy

There are four independent "how old is this update" computations. Only one has been consolidated.

| Clock | Location | State |
|---|---|---|
| `resolveMaturityClock` | `app/model/maturity-policy.ts:67-89` | source of truth, enforces the gate |
| `getRawUpdateAge` / `getRawUpdateMaturityLevel` | `app/model/container.ts:717-755` | **divergent** |
| `getContainerUpdateAge` + `getContainerMaturityLevel` | `app/api/container/update-age.ts`, `maturity-filter.ts` | **divergent** |
| client-side recompute | `ui/src/utils/container-mapper.ts` | fixed by #566 |

The two divergent ones classify against a global `DD_UI_MATURITY_THRESHOLD_DAYS` (default 7, `app/api/container/maturity-filter.ts:9`), completely independent of each container's own `maturityMinAgeDays`.

So a container configured with `maturityMinAgeDays: 30` and a 10-day-old update gets a `mature` badge and matches `?maturity=mature`, while the gate itself correctly reports it as still blocked. The badge and the gate disagree, in the same API response.

## Scope

Both are pre-existing and neither resets a soak clock on its own, so they're queued for v1.6.1 rather than blocking v1.6.0 GA.

scttbnsn:
[...]

PR #607, opened 2026-07-26
🐛 fix(maturity): stop four identity-drift bugs from silently resetting clocks and re-notifying

### LINKED ISSUE #565: Checking for updates again resets the maturity time

scttbnsn:
Four bugs, one root cause, plus the test change that unblocks the GA cut.

## The pattern

PR #568 fixed #565 by establishing what "the same update candidate" means: tag and digest, with `created` participating only when there's no digest (registries don't all derive `created` from the digest, so it drifts between scans while the candidate stays put). It fixed exactly one comparator.

There were three more places making that same judgment, each hand-rolled, each slightly different. When one of these gets it wrong there's no error and no log line, just a clock that quietly restarts or a notification that fires twice.

## What's fixed

**The maturity soak clock resets when a container is recreated.** `getResultSignature` (`app/store/container.ts`) decides whether a recreated container inherits its stashed `updateDetectedAt` / `firstSeenAt` / `maturityGatePendingSince`. It hashed `created` unconditionally, so a drifted `created` across the recreate window discarded the stash and restarted the soak — right after an update landed, which is precisely when continuity matters. #568 never touched this path.

**Notification dedup fires a duplicate `once: true` notification.** `computeResultHash` (`app/store/notification-history.ts`) hashed `suggestedTag` and `created`. Its docstring said it mirrored `hasResultChanged()` — the comparator #568 disproved. A manual "check for updates" bypasses the registry poll cache (the scheduled path passes `useRegistryPollCache`, the manual one doesn't), so it routinely returns drifted display metadata for an unchanged candidate. New hash, `hasAlreadyNotifiedForResult` reports "not notified", one-shot notification fires again.

**A finishing security scan reverts an update the watcher just found.** `runBulkScan` and `persistAndBroadcast` captured a container snapshot before the scan and wrote that whole snapshot back with only `security` changed. Scans run seconds to minutes; a watcher poll in that window could legitimately advance `result` / `updateAvailable` / `updateDetectedAt` / `maturityGatePendingSince` and have all of it silently rolled back. Next poll re-detects the same update, stamps a fresh timestamp, elapsed soak time gone. `app/security/scheduler.ts` already re-fetched at write-back; these two didn't.

**Containers with no available update get ranked and filtered as if they had one.** `getContainerUpdateAge` had no `updateAvailable` guard on its fallback, so a container gated by `maturityMode` surfaced under `?maturity=mature`, `?maturity=hot`, and `?sort=age` while its own `updateEligibility` blockers in the same response said it was blocked.

## How

`getCandidateIdentityFields` is now exported from `app/model/container.ts` as the single definition, and all three comparators derive from it. `hasCandidateIdentityChanged` and `getResultSignature` keep their exact current behavior (their existing tests pass untouched); only `computeResultHash` changes semantics. The scan paths re-fetch via `getContainerRaw` and merge only `security`, skipping the write if the container is gone.

## Upgrade note

Changing the dedup hash formula means history persisted under the old formula won't match. Users may see **one extra notification per update already in flight** immediately after upgrading. One-time, self-clearing, and called out in the CHANGELOG so it doesn't read as a new bug.

## Also here

`scripts/release-docs-identity.test.mjs` threw `RC_VERSION must end in -rc.<number>` the moment the version constant becomes a plain `1.6.0`, which would have failed the GA cut. It now branches on version shape. Both GA-branch assertions were checked against the real un-bumped rc.6 quickstart to confirm they actually fail pre-bump rather than passing vacuously.

## Verification

Every one of the 7 new regression tests was run against pre-fix source and confirmed to fail there, not just to pass here. Full app suite 12383 passing (up exactly 7 from rc.6's 12376). Full pre-push gate green: biome, qlty, qlty-smells, scripts-test, workflow-tests, typecheck-ui, 100% coverage on app and ui, both builds.

## Not in scope

Three findings from the same sweep didn't clear the blast-radius bar for a release already soaking, and are filed for v1.6.1: #604 (GHCR auth failures swallowed at debug level; two update-age clocks classifying against the global threshold instead of per-container `maturityMinAgeDays`), #605 (`agent-mismatch` shown during the agent trigger-registration window — split out of discussion #325, where the original answer was wrong), #606 (manifest `created` fetch failures swallowed to `undefined`, which is the upstream *condition* behind two of the bugs fixed here).

## Release impact

This is a runtime change, so it needs `v1.6.0-rc.7` and restarts the seven-day soak. GA is rc.7's `publishedAt` + 7 days.

## Changelog

- 🐛 Fixed identity drift across container recreation by centralizing candidate identity comparison around `tag`+`digest`, using `created` only for legacy/no-digest candidates (affects lifecycle caches and result signatures).
- 🐛 Fixed stale/incorrect notification deduplication by updating result hash semantics: identity fields derive from candidate identity helper and `created` is hashed only when no digest exists (and removed `suggestedTag` from the hash payload).
- 🐛 Fixed security scan persistence/replay issues by re-fetching the current container at write-back time and writing only security data onto the latest record.
- 🐛 Fixed concurrent watcher race conditions during on-demand security scan persistence (preserve newer watcher fields; skip/avoid resurrecting removed “zombie” containers).
- 🐛 Fixed one-shot update-candidate scan field handling by clearing update-candidate-specific scan/signature/SBOM fields when the candidate identity changes mid-scan, while preserving current-image security results.
- 🐛 Fixed maturity filtering and `sort=age` behavior by gating update-age computation and age sorting behind `updateAvailable` and excluding containers without available updates from maturity/age buckets.
- 🔧 Added/expanded regression tests covering candidate identity drift, concurrent bulk/on-demand scan persistence safety, update-age gating, container lifecycle “update-clock carry-forward”, notification hash stability, and GA release-docs identity/quickstart matrix assertions.
- 🔧 Updated bulk/on-demand security handler dependencies to support refreshed write-back via `getContainerRaw`.
- 🔧 Corrected/clarified `getContainerRaw` documentation.
- ⚠️ Breaking: Requires runtime `v1.6.0-rc.7`; restart the seven-day soak. Also requires updating all `createSecurityHandlers`/`createBulkSecurityHandlers` callers to provide `getContainerRaw`.
- 🔒 Security: Prevents security scan write-back from overwriting concurrent updates or persisting results to removed containers.

## Concerns

- Confirm every `createSecurityHandlers` and `createBulkSecurityHandlers` call site provides `getContainerRaw`.
- Ensure the production rollout includes the required runtime upgrade to `v1.6.0-rc.7` and the seven-day soak restart.
- Verify any custom store implementations correctly return `undefined` from `getContainerRaw` when a container is removed/replaced.


scttbnsn:
Went through all four. Three are real, one isn't.

**Gate ordering in `update-age.ts`: fixed.** You're right, and there's a concrete producer rather than just a theoretical stale value. `buildFallbackContainerReport` (`app/watchers/providers/docker/docker-helpers.ts:63`) spreads the container, which materializes the `updateAge` getter into a plain frozen number, then sets `updateAvailable = false` on the next line. That's exactly the shape you described. Gate moved above the cache read, with a comment explaining why the ordering matters so nobody tidies it back later.

**Update-image fields in `persistAndBroadcast`: fixed.** Good catch, and it's a problem this PR introduced. `scanUpdateImage` computes `updateScan`/`updateSignature`/`updateSbom` against the pre-scan `container.result.tag`, and re-fetching the live record is what split them apart. Before this change the snapshot and the patch were at least consistently stale together. `persistAndBroadcast` now compares candidates via `hasCandidateIdentityChanged` and clears those three keys when the candidate moved, since whatever is already stored describes the superseded candidate too. Current-image `scan`/`signature`/`sbom` still apply, because they describe the running image rather than the candidate.

**GA rc-row check scoping: fixed.** Adopted as suggested. The comment above it already claimed "for this release line" while the regex wasn't scoped to one, which is the same code-drifts-from-its-docstring problem this whole PR is about. Re-checked the scoped version against the real un-bumped quickstart to confirm it still fails pre-bump instead of passing vacuously.

**Duplicate `const db` at `container.test.ts:4962`/`4989`: not a real finding.** Those two lines sit in separate `test()` callbacks (starting at 4952 and 4979), so they're different function scopes, not a redeclaration in one block. The module parses fine: the full app suite runs 12383 tests green, and this PR's own Test & Coverage job passed. Looks like the diff context stitched the two hunks together.

Regression tests for the first two are going in now.

Issue #623, opened 2026-07-28
Does auto-update work?


ciltocruz:
### Drydock version

1.6.0-rc.7

### What happened?

Hello.

Automatic updates aren't working for me on a remote drydock. The agents are correctly configured and both have `auto_update` enabled.

But I'm getting the message "excluded-from-allow-list".

### What did you expect?

auto update dockers.

### Steps to reproduce

-

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

scttbnsn:
Found it — real bug, not your config. The update-available rule's trigger list is only meant to scope which channels get notified, but the auto-update triggers (docker/compose/command) were being run through that same list, and they can never be on it (the UI won't let you add them). So the moment you scoped the rule to specific channels, every auto-update trigger silently failed the check — controller-side, which is why both agents stopped at once. It's been lurking since rc.1; an empty list falls back to allow-all, which is why nobody hit it until now.

Fixed in v1.6.0-rc.8, publishing today. Until you're on it: clear the update-available rule's trigger list (leave it empty) and auto-update comes back immediately.

Thanks for the report — and for confirming the agents had auto_update on, that detail is what pointed at the shared controller path.

ciltocruz:
Updated to v1.6.0-rc.8 and it works again. Thanks!

PR #625, opened 2026-07-28
🐛 fix(triggers): exempt action triggers from the update-available rule trigger allow-list

scttbnsn:
## Problem

Scoping the `update-available` notification rule to specific channels — routing update alerts to just Slack, say — silently disables auto-update for the entire fleet.

The rule's trigger allow-list is meant to scope which *notification* channels get messaged. But `getUpdateAvailableAutoTriggerDispatchDecision` runs the same allow-list membership check for *every* registered trigger, including the action triggers (`docker`, `dockercompose`, `command`) that execute updates. Action-trigger ids are deliberately barred from that list on three fronts: the API validator rejects them ("Unsupported notification triggers"), the UI picker filters them out and strips any pre-existing ones on save, and the docs state action triggers "cannot be assigned to notification rules." So the moment the list is non-empty, every action trigger fails the membership check with `excluded-from-allow-list` — a decision made once, centrally, on the controller, which is why it hits local and agent-hosted triggers (`AgentTrigger` inherits the same base-class gate) simultaneously. The only evidence is a debug log line; the Update Status panel shows the container as fully eligible.

This is #623. Dormant since the allow-list landed in rc.1 (default empty list falls back to allow-all); not a regression in rc.7.

## Fix

Action-category triggers (via `getTriggerCategoryForType`, so `command` and agent-prefixed ids are covered) now bypass the allow-list *membership* check in `getUpdateAvailableAutoTriggerDispatchDecision`, returning a new `action-trigger-exempt-from-allow-list` dispatch reason. This mirrors the exemption the lifecycle-notification path has always had (`dispatchContainerForEvent`). Disabling the rule itself still blocks action triggers — `rule-disabled` remains the global kill switch, and every other real gate (auto mode, category-scoped include/exclude labels, threshold, global update mode) is untouched.

Root cause was confirmed by three independent traces (solo deep-dive plus a 4-lens multi-agent sweep with adversarial refutation) converging on the same chain before any code was written.

## Tests

- docker action trigger dispatches when the allow-list contains only notification ids
- command action trigger same (category-based, not type-list-based)
- agent-prefixed (3-part id) docker trigger same — exemption keys on category, not id shape
- disabled update-available rule still blocks action triggers
- notification trigger absent from a non-empty allow-list remains excluded (unchanged behavior)

Full app suite: 394 files / 12,400 tests, 100% coverage. Full lefthook pre-push gate green.

Follow-up (not in this PR): surface the remaining legitimate block (`rule-disabled`) as an `updateEligibility` blocker + docs Reasons-table row so operators don't need debug logs to see it.

Refs #623

## Changelog

- 🐛 Fixed `update-available` routing so **action-category triggers** bypass the notification trigger allow-list membership check.
- 🔧 Changed dispatch decisioning for the `update-available` rule when disabled due to `excluded-from-allow-list` to return `enabled: true` with reason `action-trigger-exempt-from-allow-list`.
- ✨ Added a new dispatch-reason literal `action-trigger-exempt-from-allow-list` to `NotificationRuleDispatchReason`.
- ✨ Added regression tests for `docker`, `dockercompose`/command-style, agent-prefixed action triggers, disabled-rule behavior, and excluded-notification scenarios.

## Concerns / Issues to Address

- Verify the exemption applies **only** for `excluded-from-allow-list` (and does not bypass the kill switch for other disable reasons).
- Ensure logs/debug output cover both disabled and excluded-notification cases with the correct exclusion reason.
- Confirm agent-prefixed trigger handling relies on provider/category routing as intended for all relevant action trigger ID shapes.


scttbnsn:
@coderabbitai full review

PR #627, opened 2026-07-28
📝 docs(triggers): state the action-trigger exemption from notification-rule trigger lists

scttbnsn:
Closes the documentation gap from #625: the notification-rules callout said action triggers can't be assigned to rules, but nothing explained what a scoped trigger list means for them. The callout now states the list scopes notification delivery only, action triggers are exempt and keep executing updates however the list is scoped (including agent-hosted triggers), and disabling the rule itself is the kill switch.

Docs-only; no runtime changes.

Refs #623


scttbnsn:
@coderabbitai full review

PR #641, opened 2026-07-30
✨ feat(watcher): warn on minute-precise maintenance windows + link eligibility docs from the confirm dialog

scttbnsn:
Follow-up to #640 / [Discussion #639](https://github.com/CodesWhat/drydock/discussions/639). The docs PR fixed the wrong examples; this fixes the trap at the source and makes the dialog self-explanatory.

**Startup warning for minute-precise maintenance windows.** `MAINTENANCE_WINDOW` is matched minute by minute, so `0 2-6 * * *` opens the window for one minute per matching hour, not the range the old docs promised. The Docker watcher now logs a one-time warning at init when the window cron's minute field has no `*`, with the fix in the message (`* 2-3 * * *`). Step values (`*/5`) aren't flagged. Pure check lives in `hasNarrowMinuteField()` in `maintenance.ts`; warning only, behavior unchanged.

**"Learn more" link in the update confirm dialog.** When the confirm dialog lists soft blockers ("This update is currently policy-blocked…"), it now links to the [reasons reference](https://getdrydock.com/docs/configuration/actions/update-eligibility#reasons-reference), same target the Update Status panel already uses. Our own `ConfirmDialog.vue` renders it as a real anchor via a new optional `link` field on `ConfirmOptions` — no HTML in i18n strings. New `softBlockerLearnMore` key added to all 17 locales (English placeholder, Crowdin picks it up).

Tests: `hasNarrowMinuteField` units + watcher init warn/no-warn cases; dialog link render/absent + both confirm call sites. Full app (12434) and ui (4305) suites green at 100% coverage.

## Changelog

- ✨ Added startup warnings for Docker maintenance windows with fixed minute fields.
- ✨ Added optional “Learn more” links to update confirmation dialogs when soft blockers apply.
- ✨ Added `ConfirmDialogLink`, eligibility documentation export, translations, and test coverage.
- 🔧 Updated Docker watcher mocks and update confirmation call sites.
- 🔧 Documented both changes in `CHANGELOG.md`.

## Concerns

- Verify all locale values are intentionally left as `"Learn more"` rather than translated.
- Confirm cron minute-field detection handles all supported expression formats.


scttbnsn:
@coderabbitai review

PR #662, opened 2026-08-04
🐛 fix(api): accept anonymous-auth sessions on WebSocket log streams

### LINKED ISSUE #636: WebSocket log streams reject anonymous-auth sessions (DD_ANONYMOUS_AUTH_CONFIRM)
Under `DD_ANONYMOUS_AUTH_CONFIRM=true`, every REST endpoint works (permissive middleware), but both WebSocket upgrade paths — `/api/v1/containers/:id/logs/stream` (app/api/container/log-stream.ts:602) and the system log stream (app/api/log-stream.ts:258) — gate on `isAuthenticatedSession()` (app/api/ws-upgrade-utils.ts:155), which requires `session.passport.user`. `passport-anonymous` never sets that, so anonymous deployments can never open a log stream.

Not a v1.6 regression: `isAuthenticatedSession` is byte-identical in v1.5.2 and shipped with the v1.5.0 WS streaming work, so this has been broken since the feature existed. Found live during the 2026-07-28 tri-tool NAS acceptance (evidence: /volume1/docker/tri-tool-acceptance/evidence/, item 3).

Fix direction: the WS auth gate should honor the configured auth strategy the same way the HTTP middleware does (anonymous configured+confirmed → authenticated), rather than hard-requiring a passport user. Targeting v1.6.1 (runtime; deliberately kept out of the v1.6.0 GA candidate since it predates v1.6).

scttbnsn:
Fixes #636.

Pre-existing since v1.5.0: `isAuthenticatedSession` only checked `session.passport.user`, which passport-anonymous never sets, so WS log streams always 401'd under `DD_ANONYMOUS_AUTH_CONFIRM=true`.

- `registry.isAnonymousAuthenticationActive()`: true iff the registered strategy is type `anonymous` (startup-gated by the explicit confirm).
- `isAuthenticatedSession(req, { anonymousAuthActive })` — default false, byte-identical for every existing caller; both WS upgrade sites pass the registry state for the 401 gate.
- **Rate-limit keying deliberately stays on the passport-only boolean** — anonymous clients remain IP-keyed (the adversarial round caught that reusing the gate boolean would let anonymous clients rotate session IDs past identity-keyed limits).

Fail-closed behavior for basic/OIDC/etc. is untouched. Review provenance: adversarial verify round (1 finding, repaired), independent Codex review: clean. Full app suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **WebSocket log streams no longer reject anonymous-auth sessions** ([#636](https://github.com/CodesWhat/drydock/issues/636)). Both WS upgrade paths — the system log stream and the container log stream — gated on `isAuthenticatedSession()` requiring `session.passport.user`, which `passport-anonymous` never sets, so under `DD_ANONYMOUS_AUTH_CONFIRM=true` the log stream WebSocket always rejected the upgrade even though every REST endpoint worked. `isAuthenticatedSession` now also accepts the session when anonymous authentication is the registered mode.
+
 ### Security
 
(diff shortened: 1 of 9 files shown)


biggest-littlest (approved):
Anonymous gate scoped to the registered-strategy check; rate-limit keying verified unchanged for anonymous clients.

ALARGECOMPANY (approved):
LGTM.

PR #683, opened 2026-08-09
🐛 fix(ui): resolve iconify aliases and broken references in the icon bundle

scttbnsn:
## What

The icon bundle is regenerated from the locked `@iconify-json` packages during every Docker image build (`npm run build` → `npm run icons`), but `extract-icons.mjs` only looked up plain icon entries — never aliases. lucide 1.2.121 demoted `history` to an alias of `rotate-ccw-clock`, so every image built since that pin ships without `lucide:history`, and the Audit nav icon renders blank in the Lucide icon theme (the offline iconify API module means there's no network fallback). The committed bundle still had the icon only because it was generated from stale node_modules (lucide 1.2.107).

Regenerating with the lockfile-correct packages also exposed references that have **never** resolved — broken in production all along:

- `iconoir:history`, `iconoir:gitlab`, `iconoir:stack` don't exist in the locked iconoir 1.2.11
- `fa6-brands:github/gitlab/google/microsoft` (registry icons in the Font Awesome theme) — the `@iconify-json/fa6-brands` package was never a dependency

## Fix

- `extract-icons.mjs` resolves alias chains (parent-following, depth-capped); aliases carrying rotate/flip transforms are refused with a warning since the body-only bundle can't represent them
- `iconoir:history` → `iconoir:clock-rotate-right`, `iconoir:gitlab` → `iconoir:gitlab-full`, `iconoir:stack` → `iconoir:multiple-pages` (group-view toggle in the containers list)
- `@iconify-json/fa6-brands` added (exact-pinned, build-time only)
- Bundle regenerated: 576/576 references extracted, zero warnings
- New guard test asserts every `icons.ts` reference exists in the committed bundle, so a collection rename fails CI instead of shipping a blank glyph

Wants to land before v1.6.0-rc.13 so the fix rides the release.

## Changelog

- ✨ Added exact-pinned `@iconify-json/fa6-brands@1.2.6` to `devDependencies`.
- ✨ Added a guard test for unresolved icon references.
- 🔧 Changed `extract-icons.mjs` to resolve chained aliases.
- 🔧 Changed Iconoir mappings to use icons in the locked collection.
- 🐛 Fixed missing icon references in the generated bundle.
- 🐛 Regenerated the bundle with 576 valid references.
- ⚠️ Alias transforms that require rotation or flipping remain unsupported.

## Concerns

- Confirm that the guard test runs in the standard UI test pipeline.
- Confirm that the alias depth limit covers all supported collection aliases.
- Review whether unsupported alias transforms should fail the build instead of only emitting warnings.


scttbnsn:
@coderabbitai review

scttbnsn:
Nitpick addressed in abdea8a0 — moved to devDependencies with the sibling collections; regenerated bundle is byte-identical.

biggest-littlest (approved):
LGTM

PR #686, opened 2026-08-09
🐛 fix(ui): resolve iconify aliases in the icon bundle (forward-port #683)

scttbnsn:
Forward-ports #683 to dev/v1.7. #680 fixed the symptom here (direct icon-name renames) but not the root cause: `ui/scripts/extract-icons.mjs` still dropped icons that iconify collections demote to aliases (`collection.aliases[name].parent`), so any future collection bump could silently ship broken icons again — that's exactly how rc.12 shipped a blank Audit icon.

What's in the port:
- `extract-icons.mjs`: `resolveIcon()` follows alias parent chains (depth-capped), merges alias overrides, refuses transform-requiring aliases with a warning.
- `ui/tests/boot/icon-bundle.spec.ts`: guard test — every `prefix:name` ref in `icons.ts` must exist in the bundle with a body.
- `iconoir:stack` → `multiple-pages` (#680 had picked `cube`; `multiple-pages` matches v1.6 and verified valid against locked iconoir 1.2.11).
- `@iconify-json/fa6-brands` 1.2.4 → 1.2.6 exact-pinned (devDependencies).
- Bundle regenerated: 576/576 refs, zero warnings, 229.8 KB; `npm run build` leaves the tree clean.

4508 UI tests, 100% coverage, full pre-push gate green.

## Changelog

- ✨ Added recursive Iconify alias resolution with parent-chain traversal, override merging, depth limits, and transform warnings.
- ✨ Added a guard test for missing or empty icon bundle entries.
- 🔧 Replaced `iconoir:cube` with `iconoir:multiple-pages` for the `stack` icon.
- 🔧 Pinned `@iconify-json/fa6-brands` to `1.2.6`.
- 🐛 Fixed extraction of indirect icon aliases.

## Concerns

- Verify aliases with rotation or flipping produce actionable warnings and do not enter the body-only bundle.
- Keep the guard test aligned with all supported `icons.ts` reference formats.
- Confirm the regenerated bundle contains all 576 references with nonempty bodies.


biggest-littlest (approved):
LGTM

PR #700, opened 2026-08-12
chore(release-cut): add auditable soak_override_reason for GA promotion

scttbnsn:
## Summary
- Adds a GA-only `soak_override_reason` workflow_dispatch input to `release-cut.yml`. The 604800s (seven-day) soak constant is unchanged and stays the strict default.
- When the candidate is under seven days old: a blank/whitespace-only reason keeps the original hard failure verbatim; a reason under 20 trimmed characters hard-fails with a new, clear error; a real reason downgrades the check to a `::warning::` and continues.
- The override is recorded in `$GITHUB_STEP_SUMMARY` (candidate tag, age, reason) and appended to the published release notes body.
- `soak_override_reason` is rejected outright on a prerelease cut, same as `candidate_tag`/`candidate_digest`.
- The reason is passed via `env:` and referenced only as `"${SOAK_OVERRIDE_REASON}"` inside `run:` — never templated directly — to avoid a zizmor template-injection finding on this attacker-controllable input (any repo collaborator can dispatch the workflow).

Context: v1.6.0 GA needs to promote `v1.6.0-rc.13` (~3 days old, fleet soak passed 2026-08-09) ahead of the seven-day window tonight. This input is the auditable escape hatch for that, not a change to the default policy.

## Test plan
- [x] `npm run test:workflows` — 78/78 passing, including a new `.github/tests/release-cut-soak-override.test.ts` that extracts and *executes* the real `source` step script (stubbed `gh`/`date`, a real scratch git repo) covering: blank reason still fails with the unchanged message, whitespace-only reason does not open the gate, too-short reason hard-fails, a valid reason bypasses with a warning + summary + notes file, an already-7-day-old candidate needs no reason, and prerelease rejection.
- [x] `zizmor .github/workflows/` — no findings.
- [x] Full pre-push gate (lefthook: biome, qlty, scripts-test, workflow-tests, typecheck, coverage, build, zizmor) — green.
- [x] YAML parses correctly (`inputs` includes `soak_override_reason`).

## Changelog

- ✨ Added GA-only `soak_override_reason` input.
- 🔧 Changed GA soak validation:
  - Keeps the 604800-second default.
  - Rejects blank or whitespace-only reasons.
  - Rejects trimmed reasons shorter than 20 characters.
  - Downgrades valid overrides to warnings.
  - Records override details in summaries and release notes.
- 🔒 Security:
  - Passes the input through `env:` instead of shell interpolation.
  - Rejects override reasons for prerelease cuts.
- ✨ Added integration and security tests.
- 🔧 Isolated test Git repositories from inherited `GIT_DIR` variables.

### DIFF
--- a/.github/tests/release-cut-retry-workflow.test.ts
+++ b/.github/tests/release-cut-retry-workflow.test.ts
@@ -207,6 +207,11 @@ test('release-cut requires an exact, seven-day-old RC candidate for GA promotion
       required: false,
       type: 'string',
     },
+    soak_override_reason: {
+      required: false,
+      type: 'string',
+      default: '',
+    },
   });
   expect(sourceStep?.id).toBe('source');
   expect(sourceStep?.run).toContain('publishedAt');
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
Fixed in 63b605d4 — the reason no longer reaches stdout. The `::warning::` now carries a fixed message and points at the job summary and release notes, which still record the full text. Added a regression case feeding a multiline reason containing `::add-mask::` and `::error::` lines and asserting neither shows up in stdout. 79/79 workflow tests pass.

biggest-littlest (approved):
Auditable override with the constant untouched. Good catch on keeping the reason out of the workflow command.

ALARGECOMPANY (approved):
LGTM.

PR #701, opened 2026-08-12
🐛 fix(web): read star history through GraphQL so the chart renders

scttbnsn:
## What

The README star-history chart has been stuck on the placeholder ("loading") SVG. Root cause: GitHub's REST `/stargazers` endpoint now returns **401 anonymously**, and the route treated `GITHUB_TOKEN` as optional — so `fetchStarredTimestamps()` always returned `undefined` and the route permanently served `renderStarHistoryFallbackSvg`.

Handing it a token doesn't fix it either: for a fine-grained token the endpoint reports `x-accepted-github-permissions: metadata=read; contents=write`, and it 403s with `metadata=read` alone. Granting `contents=write` to a public star chart is not a trade worth making.

## Fix

Read the stargazers through GraphQL instead. That connection needs only `metadata=read`, returns `starredAt` directly, and pages with a cursor. Verified live against `CodesWhat/drydock`:

```
{"data":{"repository":{"stargazerCount":216,"stargazers":{"edges":[{"starredAt":"2026-02-10T16:53:13Z"},...]}}}}
```

Behaviour that did not change: every incomplete outcome still returns `undefined` and falls back rather than rendering a truncated total, the whole pagination run shares one `FETCH_DEADLINE_MS` deadline, and the success/failure cache headers are untouched. New: no token means fall back, since GraphQL has no anonymous mode.

## Verification

- `npx tsc --noEmit` clean.
- `npm run test:scripts --prefix apps/web` — 57/57. Two assertions pinned the REST pagination shape, so they're repointed at the GraphQL shape; added a case covering the endpoint swap and the no-token fallback.
- Ran the route against a real token on a local dev server: `200`, 7,578 bytes, real 216-star curve rendered — not the fallback.

`GITHUB_TOKEN` is set on the `drydock-website` Vercel project (production + preview) from a new `drydock-star-history` fine-grained token scoped to `CodesWhat/drydock` with Metadata: Read-only.

## Changelog

- ✨ Added GraphQL stargazer queries with `starredAt` ordering and cursor pagination.
- 🔧 Changed the star-history route from REST `/stargazers` to GraphQL.
- 🔧 Changed authentication to require `GITHUB_TOKEN` with `metadata: read`.
- 🐛 Fixed fallback handling for missing tokens, invalid responses, incomplete pagination, and pagination limits.
- 🔒 Removed dependence on the REST endpoint, which requires `contents: write`.
- ✨ Added tests for GraphQL requests, pagination, REST endpoint exclusion, and no-token fallback.

## Concerns

- Confirm the pagination limit matches the maximum expected repository star count.
- Complete the `coderabbitai` review after rate limiting ends.

### DIFF
--- a/apps/web/scripts/marketing-performance.test.mjs
+++ b/apps/web/scripts/marketing-performance.test.mjs
@@ -71,7 +71,9 @@ test("star history route never renders partial stargazer data and bounds its fet
 test("star history route reads stars through GraphQL, not the REST stargazers endpoint", () => {
   // REST /stargazers now 401s anonymously and demands contents=write from a
   // fine-grained token; the GraphQL connection needs only metadata=read.
-  assert.match(starHistoryRouteSource, /https:\/\/api\.github\.com\/graphql/u);
+  // A substring check, not a regex: an unanchored URL pattern reads as host
+  // matching to CodeQL, and this only asserts the endpoint appears in source.
+  assert.ok(starHistoryRouteSource.includes('"https://api.github.com/graphql"'));
   assert.equal(starHistoryRouteSource.indexOf("/stargazers?per_page="), -1);
   assert.match(starHistoryRouteSource, /orderBy:\{field:STARRED_AT,direction:ASC\}/u);
   // No token means no data at all, so the route must fall back rather than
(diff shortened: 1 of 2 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
Fixed in a949996c. `hasNextPage` must now be an actual boolean, any edge without a parseable `starredAt` fails the whole run, and the series returns only on `hasNextPage === false`. Tests updated to assert the strict shape instead of the permissive one. Re-verified live against the real token: 200, real 216-star curve, no fallback.

The CodeQL anchor alert on the test file was fixed in 88095e8c — that assertion is a substring check now, and the `CodeQL` check is green.

biggest-littlest (approved):
GraphQL swap plus the strict page validation. Verified the chart renders.

ALARGECOMPANY (approved):
LGTM.

Issue #704, opened 2026-08-12
IFTTT trigger sends its webhook with no request timeout


scttbnsn:
`app/triggers/providers/ifttt/Ifttt.ts:84` calls `axios(options)` with no `timeout`, so an unresponsive `maker.ifttt.com` hangs the notification until Node's socket default gives up.

Its nine axios-calling siblings (Apprise, Discord, Google Chat, Matrix, Mattermost, ntfy, Rocket.Chat, Teams, Telegram) all pass `getOutboundHttpTimeoutMs()` from `app/configuration/runtime-defaults.js`. IFTTT is the only one that doesn't.

Not an SSRF issue: the URL is the fixed `maker.ifttt.com` endpoint, not operator-supplied. It's a liveness gap, and it's the reason [`SECURITY-ASSURANCE.md`](https://github.com/CodesWhat/drydock/blob/main/SECURITY-ASSURANCE.md) has to carry a 'no' in the timeout column for this one path.

**Fix:** import `getOutboundHttpTimeoutMs` and add `timeout: getOutboundHttpTimeoutMs(),` to the options object. `Ifttt.test.ts` has six `expect(axios).toHaveBeenCalledWith({...})` assertions that pin the exact options shape, so all six need the new key. Then flip the IFTTT row in `SECURITY-ASSURANCE.md` to 'yes' and drop the sentence pointing at this issue.

Found by CodeRabbit while reviewing #702. Kept out of that PR to keep a docs forward-port from growing a code change.

scttbnsn:
Fixed on `dev/v1.7` in #707. IFTTT now passes `getOutboundHttpTimeoutMs()` like its nine sibling providers, the six exact `toHaveBeenCalledWith` assertions carry the new key, and `SECURITY-ASSURANCE.md`'s outbound-controls table reads `yes` for the IFTTT timeout column.

Closing by hand because the `Fixes:` footer only auto-closes on a merge to the default branch, and this landed on the dev branch. It ships with v1.7.0.

PR #707, opened 2026-08-12
fix(ifttt): apply the shared outbound HTTP timeout

scttbnsn:
IFTTT was the one outbound path in the codebase with no timeout at all. It called `axios(options)` bare while its nine sibling providers all pass `getOutboundHttpTimeoutMs()`, so a hung IFTTT endpoint hung the notification indefinitely.

Three parts:

- `Ifttt.ts` now passes the shared timeout the same way its siblings do.
- The six `toHaveBeenCalledWith({...})` assertions in `Ifttt.test.ts` carry the new key. Left exact rather than loosened to `objectContaining` — the exactness is what makes them worth having.
- `SECURITY-ASSURANCE.md`: the IFTTT row's timeout column reads `yes`, and the sentence naming IFTTT as the untimed path is gone, since it stops being true when this lands.

Fixes: #704

## Changelog

- ✨ Added `timeout: getOutboundHttpTimeoutMs()` to IFTTT webhook requests.
- 🔧 Changed six `Ifttt.test.ts` Axios option assertions to expect a `30000` ms timeout.
- 🔒 Updated `SECURITY-ASSURANCE.md` to mark IFTTT timeout support as enabled.
- 🗑️ Removed the outdated statement that IFTTT lacks a timeout and the related issue reference.

## Concerns

- Verify all IFTTT Axios calls use the shared outbound timeout.
- Verify the test expectation remains aligned with the configured default timeout.

### DIFF
--- a/SECURITY-ASSURANCE.md
+++ b/SECURITY-ASSURANCE.md
@@ -97,7 +97,7 @@ individually rather than claiming a blanket policy:
 | Icon fetch (`app/api/icons/`) | yes | no | yes | no |
 | Release notes (`app/release-notes/`) | yes | no | no | no |
 | Notification providers calling axios directly (Apprise, Discord, Google Chat, Matrix, Mattermost, ntfy, Rocket.Chat, Teams, Telegram) | yes | no | no | no |
-| IFTTT notification provider (`app/triggers/providers/ifttt/`) | no | no | no | no |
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Approved.

ALARGECOMPANY (approved):
Approved.

Issue #556, opened 2026-07-17
I believe that the maturity days are not applied as they should be


ciltocruz:
### Drydock version

1.6.0-rc.1

### What happened?

I have a container configured with a 5-day maturity period.

<img width="374" height="399" alt="Image" src="https://github.com/user-attachments/assets/134cd7be-9c2a-479a-b7eb-f2b4026c8cc3" />

I expect the system to allow an update if the new image is older than that.

But no, it seems to still be based on the date Drydock discovers the new image.

<img width="355" height="311" alt="Image" src="https://github.com/user-attachments/assets/45fc0559-f3cd-4356-a25e-c55d033d455f" />

### What did you expect?

I expect the system to allow an update if the new image is older than that.

### Steps to reproduce

1. I have a Docker container with an update from weeks ago in GHCR, for example.
2. I set the maturity to 5 days (only mature).
3. Should I start DryDock from scratch? -> (It seems that with each DryDock image recreation, the image discovery time is reset.)
4. The image cannot be updated, even though the latest tag, version, or version number is older than the required maturity.

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

scttbnsn:
First off, thank you for all the testing you've been doing on these RCs. Between this one, #496, #535, #490 and the discussions, you've personally flushed out more real bugs in the maturity/update pipeline than anyone else, and every report has been detailed enough to actually act on. Genuinely appreciated, this kind of feedback is what release candidates are for.

Dug into this fully and found the actual mechanism. It's not what either of us guessed at first, and it's two separate problems that compound.

**1. GHCR publish dates never load without a token.** Pulling public images from GHCR needs no credentials, which is the setup our docs describe. But the publish date lookup uses a different API (GitHub's Packages REST API), and that one returns 401 without a `read:packages` token even for fully public packages. Right now that failure is swallowed silently (it only shows at debug log level), so the maturity clock quietly falls back to "when drydock first saw the update" instead of the image's real publish date. That's exactly your symptom, and with a tokenless GHCR image it happens every time. Docker Hub isn't affected, its metadata API works anonymously. This is GHCR-specific (and lscr.io, which shares the same code path).

[...]

PR #743, opened 2026-08-16
fix(store): persist updatePolicyRetentionCache across process restart

scttbnsn:
## Summary
- Round 3 of the #565 thread: a controller self-update (recreate -> SIGTERM -> new process) still wiped stashed maturity-policy overrides for a replacement container, even though the sibling lifecycle/clock cache (`updateLifecycleCache`) already survives the same restart per #556.
- `updatePolicyRetentionCache` only ever lived in a bare in-memory `Map`, so `stashUpdatePolicyForReplacement()`'s stash was gone before the replacement container's `insertContainer()` call could consume it in the new process.
- Gives `updatePolicyRetentionCache` the same durable, write-through-backed collection `updateLifecycleCache` already has: a new `app/store/update-policy-retention-cache.ts` module, write-through on stash, delete-on-consume, and `rehydrateUpdatePolicyRetentionCacheFromStore()` wired into `store/index.ts` startup before the first `insertContainer` runs.
- Does not touch the round-1 (clock-restart, #568) or round-2 (empty-override-authoritative, #624) fixes, which are already correct.
- The remaining half of the report — policy loss from a pure 0-container remote-agent handshake with no controller restart — isn't explained by this fix. The existing zero-container handshake path already skips `pruneOldContainers`/`deleteContainer` entirely, so that part needs more repro detail from the reporter before a second root cause can be pinned down.

## Test plan
- [x] `npx vitest run store/container.test.ts store/index.test.ts store/update-lifecycle-cache.test.ts store/update-policy-retention-cache.test.ts` — new "simulated process restart" test pins the fix; also new dedicated unit tests for the new store module and wiring-order assertions in `store/index.test.ts`
- [x] `npm test` in `app/` — 12907 tests pass, 100% coverage
- [x] `npx tsc --noEmit` and `npx biome check .` clean
- [x] Pre-push hooks (biome, qlty, workflow-tests, typecheck-ui, coverage, build) all green

Fixes: #565

## Changelog

- ✨ Added a durable LokiJS-backed store for update-policy retention records.
- 🔧 Changed `updatePolicyRetentionCache` to write through to persistent storage.
- 🔧 Changed cache consumption, expiration, and size eviction to delete persistent records.
- 🔧 Changed startup to rehydrate unexpired records before the first `insertContainer` call.
- 🐛 Fixed maturity-policy overrides being lost during controller process restarts.
- ✨ Added tests for persistence, rehydration, expiration pruning, eviction, deletion, and startup ordering.

## Concerns

- Verify behavior during zero-container remote-agent handshakes without a controller restart. This PR does not cover that loss path.
- Verify the new store follows existing store initialization and persistence conventions.
- Review whether `unknown` for `updatePolicyOverrides` preserves the required type safety.
- Check whether cache persistence and rehydration logic duplicates existing store utilities.

### DIFF
--- a/app/store/container.test.ts
+++ b/app/store/container.test.ts
@@ -6,6 +6,7 @@ import { updateContainerFromInspect } from '../watchers/providers/docker/contain
 import { pruneOldContainers } from '../watchers/providers/docker/container-init.js';
 import * as container from './container.js';
 import * as updateLifecycleCacheStore from './update-lifecycle-cache.js';
+import * as updatePolicyRetentionCacheStore from './update-policy-retention-cache.js';
 
 vi.mock('./migrate');
 vi.mock('../event');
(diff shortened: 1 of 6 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Root cause adversarially verified (bare-Map retention cache vs the durable lifecycle cache from #556); fix mirrors the shipped persistence pattern and the restart regression test pins it.

ALARGECOMPANY (approved):
12907 tests at 100% coverage, CodeRabbit clean, qlty failures are the known billing non-gate.

PR #745, opened 2026-08-16
ci(verify): remove legacy security-actions bridge

scttbnsn:
## Summary
- Ruleset 13077055 was narrowed to require `Security: Actions / Workflow Security` (the reusable go-ci caller's composite context) instead of the old plain `Security: Actions` context. The `legacy-security-actions` fail-closed bridge job in `.github/workflows/ci-verify.yml`, which existed only to republish the old plain name, is now unreferenced by branch protection and has been removed.
- Updated `.github/tests/ci-verify-workflow.test.ts` to drop the assertion on the removed job's name and add a retired-context rejection test (mirrors portwing #122 / sockguard #260): asserts the `legacy-security-actions` job is gone and that no job (other than the legitimate `security-actions` reusable caller, whose `name:` field is the composite prefix, not a standalone check) republishes the bare `Security: Actions` name. The regex is anchored to `^[ \t]*name:` with optional quotes so it can't false-positive on an input key like `workflow-security-check-name:`.
- Updated `.github/tests/harden-runner-workflow.test.ts` to drop the block-egress carve-out and dedicated bridge test that only existed for the removed job; every remaining Harden Runner step now uses the standard `audit` policy.

## Test plan
- [x] `npm run test:workflows` — 167/167 pass
- [x] Full `git push` pre-push gate (clean-tree, ts-nocheck, biome, qlty, qlty-smells, scripts-test, workflow-tests, typecheck-ui, coverage, build, zizmor) — all green

## Changelog

- 🗑️ Removed the obsolete `legacy-security-actions` bridge job from `.github/workflows/ci-verify.yml`.
- 🔧 Updated workflow tests to reject the retired job and bare `Security: Actions` check context.
- 🔒 Removed the Harden Runner egress-policy exception for the bridge job.
- 🔧 Removed the bridge-specific Harden Runner test.
- 🔧 Verified workflow tests and the full pre-push gate pass.

### DIFF
--- a/.github/tests/ci-verify-workflow.test.ts
+++ b/.github/tests/ci-verify-workflow.test.ts
@@ -45,11 +45,36 @@ function getTestJobStep(name: string): WorkflowStep | undefined {
 test('required ci-verify jobs publish stable plain-text check names', () => {
   const workflow = loadWorkflow();
 
-  expect(workflow.jobs?.['legacy-security-actions']?.name).toBe('Security: Actions');
   expect(workflow.jobs?.lint?.name).toBe('Quality: Lint');
   expect(workflow.jobs?.test?.name).toBe('Quality: Test & Coverage');
   expect(workflow.jobs?.build?.name).toBe('Build');
   expect(workflow.jobs?.e2e?.name).toBe('E2E: Cucumber');
+});
+
+test('the retired "legacy-security-actions" bridge job and its plain check name stay removed', () => {
+  const workflow = loadWorkflow();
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Bridge unreferenced after the ruleset narrow; retired-context rejection test pins the old name out.

ALARGECOMPANY (approved):
All six required contexts green, legacy context absent from the rollup as intended.

PR #756, opened 2026-08-17
feat(updates): gate admission through the action policy resolver

scttbnsn:
Slice 3 of the 6.0.1 per-action policy package (#511), on top of #752/#755.

- `resolveUpdateTrigger` resolves the update trigger through `selectActionTrigger` (policy-aware, specificity-ranked) instead of first-compatible-by-insertion-order. 404 semantics preserved.
- Defense-in-depth guard in `prepareContainerUpdateRequest`: `source === 'automatic'` requests are rejected 409 unless the resolved policy state for the trigger is `auto`. Applies to both explicit-trigger and auto-resolved paths, so automatic dispatch can never execute through a trigger the policy only grants `manual`. Manual/API admission is unchanged (admits `manual` and `auto`).
- Global `updateMode` gate and hard-blocker check untouched.
- Tests: the automatic-source guard matrix (auto admits, manual rejects 409, legacy `AUTO` bool mapping preserved), regression-matrix case 10 (same-name containers on different agents admit independently), and `getId()` backfill on trigger mocks in the admission-path consumers (webhook, container-actions, dependency-groups).

Auto-dispatch wiring (the winner-selection gate inside `Trigger.ts` plus the double-dispatch closure) is slice 4.

### ✨ Added
- Added `triggerTypes` filtering to `selectActionTrigger`.
- Added automatic-source admission tests for default and legacy `AUTO` policies.
- Added same-name container tests across agents.
- Added `getId()` support to trigger mocks and trigger-path consumers.

### 🔧 Changed
- Resolve update triggers with the policy-aware, specificity-ranked action policy resolver.
- Filter trigger candidates by `options.triggerTypes` before ranking.
- Require resolved triggers to provide `getId()`.
- Reject automatic-source requests with HTTP 409 unless the resolved trigger policy state is `auto`.
- Preserve manual/API admission behavior for `manual` and `auto` states.
- Preserve the global `updateMode` gate and hard-blocker check.
- Widen the trigger type guard parameter to `unknown` without changing runtime checks.

### 🐛 Fixed
- Prevent automatic admission from creating operations or executing triggers when the resolved trigger policy is not `auto`.
- Preserve existing 404 behavior.

## Concerns

- Auto-dispatch winner selection remains deferred to slice 4.
- Double-dispatch handling remains deferred to slice 4.
- Replace repeated inline `getId()` trigger mocks with a shared test factory or utility.

### DIFF
--- a/app/api/container-actions.test.ts
+++ b/app/api/container-actions.test.ts
@@ -581,7 +581,7 @@ describe('Container Actions Router', () => {
         .mockReturnValueOnce(clearedContainer); // after updateContainer clears flag
       mockUpdateContainer.mockReturnValue(clearedContainer);
       const mockTriggerFn = vi.fn().mockResolvedValue(undefined);
-      const trigger = { type: 'docker', trigger: mockTriggerFn };
+      const trigger = { type: 'docker', trigger: mockTriggerFn, getId: () => 'docker.default' };
       mockGetState.mockReturnValue({ trigger: { 'docker.default': trigger } });
 
       const handler = getHandler('post', '/:id/update');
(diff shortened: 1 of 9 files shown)


scttbnsn:
All three review findings addressed in bc3f824d: the type guard now requires getId() so a malformed explicit trigger fails with the existing 500 instead of a TypeError, selectActionTrigger honors triggerTypes as a pre-ranking candidate filter, and the explicit-trigger automatic-source 409 test is in. e7369284 widens the guard param to unknown, which cleared the CodeQL null-comparison alert. E2E Playwright failure was flake, rerun green.

biggest-littlest (approved):
Reviewed the admission wiring and the review-fix commits. LGTM.

PR #829, opened 2026-08-21
fix(deps): revert e2e js-yaml override to v3 for artillery compatibility

scttbnsn:
## What

The rc.2 release cut (run 32444361346) failed because main's push-triggered CI Verify run (32444352693) died in both Artillery load-test jobs:

```
SyntaxError: The requested module 'js-yaml' does not provide an export named 'default'
```

#771 bumped the `js-yaml` npm **override** in `e2e/package.json` from 3.15.1 to 5.3.0. Overrides are global, so that forced ESM-only js-yaml 5 onto artillery's internals, which use a default import and `safeLoad` (both gone in v4+/v5). The load-test jobs only run on main pushes and workflow_dispatch, never on PRs, so #771 merged green and the break surfaced on the promotion push.

## Changes

- Revert the override to 3.15.1 (artillery's own range is `^3.15.1`).
- Renovate exception scoped to `e2e/package.json` only: no major js-yaml bumps until artillery adopts the v4+ API. apps/web stays on js-yaml 5, which its build handles fine.
- New PR-gating regression test `e2e/tests/artillery-startup.test.js` (runs in the Cucumber job's `test:support`): one VU of the real ci scenario against a dead target, exercising artillery's full startup path. Verified **red** on the js-yaml 5 override, **green** on v3.

## Verification

- `npx artillery run test/test.yml -e ci` with a 1-VU phase: parses and executes, only ECONNREFUSED from the dead target.
- Codex review: two findings, both addressed (regression test added; renovate description completed). Lockfile/scoping checks passed.

After merge: dispatch ci-verify.yml on dev/v1.7 (dispatch runs the load-test jobs) to confirm green, then re-promote to main and re-dispatch the rc.2 cut.

## Changelog

- 🐛 Fixed Artillery compatibility by pinning `e2e/package.json` to `js-yaml` `3.15.1`.
- 🔧 Changed Renovate rules to block major `js-yaml` updates only for `e2e/package.json`.
- ✨ Added a regression test for Artillery startup and YAML loading.

## Concerns

- Confirm the new test is included in the PR-gating test command.
- Verify that the test asserts the expected Artillery phase log and rejects module-loading or syntax errors.
- Keep `apps/web` on `js-yaml` `5.x`.

### DIFF
--- a/e2e/package-lock.json
+++ b/e2e/package-lock.json
@@ -2575,13 +2575,6 @@
       ],
       "license": "MIT"
     },
-    "node_modules/argparse": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
-      "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
-      "dev": true,
-      "license": "Python-2.0"
-    },
     "node_modules/arrivals": {
       "version": "2.1.2",
       "resolved": "https://registry.npmjs.org/arrivals/-/arrivals-2.1.2.tgz",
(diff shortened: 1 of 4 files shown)


biggest-littlest (approved):
Reviewed the override revert and the PR-gating artillery startup test. Verified red/green evidence in the PR body.

ALARGECOMPANY (approved):
Scoped renovate exception looks right; apps/web unaffected.

PR #835, opened 2026-08-21
chore(config): sync biome.json schema to installed 2.5.8

scttbnsn:
biome.json's `$schema` said 2.5.5 while the root package.json pins @biomejs/biome 2.5.8, violating the documented constraint that they match. Single biome.json at root governs all workspaces (verified: no per-workspace configs). Lint re-run green from root and app/, ui/, e2e/, apps/web under the corrected schema.

Note for a future PR: biome 2.5.8 emits an info-level deprecation for `linter.recommended` (superseded by `preset`); non-fatal, deliberately not migrated here.


biggest-littlest (approved):
Schema now matches the pinned 2.5.8 CLI; lint verified green in all five workspaces.

ALARGECOMPANY (approved):
Single root biome.json confirmed; no per-workspace configs missed.

PR #836, opened 2026-08-21
fix(e2e): raise flaky Playwright wait ceilings past the app's own budgets

### LINKED ISSUE #832: flaky: two Playwright specs failed 3 CI runs this week, all passed on rerun
Three Playwright failures this week, all on unchanged code, all green on rerun. Two distinct specs:

1. `playwright/dashboard.spec.ts` > "update operation should reach a terminal status". Flaked on #817 and #827 PR runs.
2. `playwright/auth.setup.ts` > setup failed with "Drydock startup watcher scans did not produce the complete QA fixture snapshot". Flaked on the main push run after the rc.2 promotion (run 32451560076), which failed the second release-cut attempt and cost a full cut cycle.

Both smell like startup/timing races in the QA fixture environment rather than assertion bugs: the auth setup one is explicitly the watcher scan not finishing before the snapshot check, and the dashboard one waits on an update reaching a terminal state.

Worth a look before GA week since a flake on a main push run blocks the release cut's CI wait, not just a PR:

- [ ] auth.setup: poll/retry the fixture snapshot check instead of a single-shot assertion, or raise its wait ceiling
- [ ] dashboard terminal-status: check what the actual timeout budget is vs. how long the QA container takes under CI load
- [ ] if either reproduces locally under `--repeat-each`, pin down the race instead of padding timeouts

scttbnsn:
Fixes #832. Both flakes were waits budgeted tighter than the backend operations they cover, so under CI load the test raced the app's own state machine:

- **auth.setup.ts** fixture-snapshot poll: 180s ceiling vs one full startup scan across the 31 watched qa-compose containers. Now 240s poll / 300s setup timeout (same 60s buffer as before). This is the flake that failed the rc.2 promotion's main-push run and cost a cut cycle.
- **dashboard.spec.ts** terminal-status poll: 60s budget vs a pipeline whose health-gate phase alone is allowed `NON_SELF_UPDATE_HEALTH_TIMEOUT_MS` = 120s. Now 150s poll / 200s test timeout. (QA runs without the security scanner, so the scan/SBOM phases don't apply; the comment says so.)

Each number is justified inline from the app constant it must outlast, not padded blindly. The e2e-playwright-workflow guard test that pins the auth.setup constants is updated in the same commit (it correctly rejected the first push). Workflow-verified: an adversarial review pass caught two inaccurate justification comments (container count, phase narration), corrected here. `playwright test --list` parses all 34 tests; e2e lint clean; workflow tests green.

## Changelog

- 🔧 Changed `auth.setup.ts`:
  - Increased fixture-snapshot polling from 180s to 240s.
  - Increased setup timeout from 240s to 300s.
  - Updated timeout rationale comments.
- 🔧 Changed `dashboard.spec.ts`:
  - Increased terminal-status polling from 60s to 150s.
  - Increased test timeout from 90s to 200s.
  - Updated timeout rationale comments.
- 🔧 Changed `.github/tests/e2e-playwright-workflow.test.ts`:
  - Updated expected authentication timeout constants.
- ✨ Added coverage for QA container startup scans and the 120s health-gate budget.

## Concerns

- Verify that the dashboard timeout change from 90s to 200s matches the intended 200s test budget.
- Confirm that timeout constants remain synchronized between `auth.setup.ts` and the workflow guard test.
- Confirm that all 34 Playwright tests parse, e2e lint passes, and workflow tests pass.

### DIFF
--- a/.github/tests/e2e-playwright-workflow.test.ts
+++ b/.github/tests/e2e-playwright-workflow.test.ts
@@ -97,8 +97,8 @@ test('Playwright waits for one complete QA scan before browser tests and parks b
   expect(authSetup).toContain('/api/v1/containers?limit=100');
   expect(authSetup).not.toContain('/api/v1/containers/watch');
   expect(authSetup).toContain('page.request.get');
-  expect(authSetup).toContain('180_000');
-  expect(authSetup).toContain('setup.setTimeout(240_000)');
+  expect(authSetup).toContain('240_000');
+  expect(authSetup).toContain('setup.setTimeout(300_000)');
   expect(authSetup).toContain('Nginx (Hooked)');
   expect(authSetup).toContain("container.labels?.['dd.group'] === 'web-stack'");
   expect(authSetup).toContain('container.result');
--- a/e2e/playwright/auth.setup.ts
+++ b/e2e/playwright/auth.setup.ts
@@ -46,7 +46,20 @@ function hasCompleteQaFixtureSnapshot(payload: unknown): boolean {
   );
 }
 
-setup.setTimeout(240_000);
+// qa-compose.yml watches 31 containers (37 services minus unwatched infra
+// like dex, mosquitto, and trivy-server) across every registry provider the
+// suite exercises (docker hub, ghcr, mirror.gcr.io, ...). initWatcher() runs
+// one full scan across all of them at boot before the cron schedule takes
+// over (DD_WATCHER_LOCAL_CRON is parked at Feb 29 specifically so no second
+// scan can race this one) — there's no per-container event to hook, only the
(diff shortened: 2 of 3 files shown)


biggest-littlest (approved):
Ceilings traced to the app's own budgets (health gate 120s, full-fleet startup scan); guard test moved with the change.

ALARGECOMPANY (approved):
Adversarial review findings (comment accuracy) confirmed fixed.

biggest-littlest (approved):
Re-approving the rebuilt branch: identical 3-file diff, now based on dev/v1.7 instead of main.

ALARGECOMPANY (approved):
Confirmed cherry-pick 45b179bc matches the previously reviewed dfe905b3 diff.

PR #841, opened 2026-08-21
chore(commits): retire gitmoji on the maintenance line to match the org convention

scttbnsn:
Ports 9abe4113 (dev/v1.7's gitmoji retirement, #696) byte-for-byte to the maintenance line, so both live branches enforce the same plain Conventional Commits rule: `<type>(scope): <description>`, 11 types, no emoji, retired custom types (`security`/`deps`/`config`/`deploy`/`remove`) rejected.

All six files were confirmed identical between the branches before the port, so this is the same change, not a re-derivation. Existing gitmoji history is untouched; only new commits are held to the new rule. Validator tests: 31/31 pass, and the commit itself was accepted by the new hook it ships.

House-audit follow-up: dev/v1.6 was the last branch still enforcing the retired convention.

## Changelog

🔧 **Changed**
- Replaced Gitmoji-prefixed commits with plain Conventional Commits.
- Enforced `<type>(scope): <description>` format.
- Added breaking-change syntax and `fixup!`/`squash!` exemptions.
- Updated documentation, CI-generated messages, validator tests, and range fixtures.
- Allowed 11 Conventional Commit types.

🗑️ **Removed**
- Rejected `security`, `deps`, `config`, `deploy`, and `remove` types.
- Rejected leading emojis and legacy type aliases.

⚠️ **Breaking**
- New commits must use plain Conventional Commits.
- Existing Gitmoji history remains valid.

## Concerns

- Verify both live branches use the same validator rules and documentation.

### DIFF
--- a/.github/workflows/i18n-crowdin.yml
+++ b/.github/workflows/i18n-crowdin.yml
@@ -89,11 +89,11 @@ jobs:
           download_translations: true
           localization_branch_name: l10n_crowdin
           create_pull_request: true
-          pull_request_title: "🔧 chore(i18n): sync translations from Crowdin"
+          pull_request_title: "chore(i18n): sync translations from Crowdin"
           pull_request_body: "Automated sync from Crowdin. Review wording, then merge."
           pull_request_labels: "translations,crowdin"
           pull_request_base_branch_name: ${{ steps.base.outputs.name }}
-          commit_message: "🔧 chore(i18n): sync translations from Crowdin"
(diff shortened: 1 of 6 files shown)


biggest-littlest (approved):
Diff matches 9abe4113 on dev/v1.7; validator tests pass.

ALARGECOMPANY (approved):
Same change as the dev/v1.7 retirement, ported cleanly.

PR #842, opened 2026-08-21
fix(security): fail-closed on ZAP warnings and zizmor skip

scttbnsn:
Two house-audit findings, both the same shape: a gate that silently passes when its tool doesn't run.

- Removed `cmd_options: '-I'` from all three ZAP jobs (ci-verify's dast-zap-baseline, security-dast-web's full scan and demo baseline). `-I` told ZAP to ignore warnings entirely, so the scans could never fail on findings. `.zap/rules.tsv` already exists for targeted suppressions (one IGNORE entry today), which is the right mechanism — suppress named findings, not the whole signal.
- Lefthook's pre-push zizmor step now fails with an install hint when the binary is missing instead of silently skipping. Same shape sockguard already uses.

Workflow tests: 208/208 pass. Ran the new zizmor command locally: no findings, exit 0. If ZAP surfaces real warnings on this PR's DAST runs, they're triaged as findings, not re-suppressed wholesale.

### DIFF
--- a/.github/workflows/ci-verify.yml
+++ b/.github/workflows/ci-verify.yml
@@ -951,7 +951,6 @@ jobs:
           rules_file_name: .zap/rules.tsv
           allow_issue_writing: false
           fail_action: true
-          cmd_options: '-I'
 
       - name: Convert ZAP JSON report to SARIF
         if: always() && hashFiles('report_json.json') != ''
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
ZAP -I removal checked against all three jobs; rules.tsv stays as the suppression path. Zizmor step matches sockguard's fail-loud shape.

ALARGECOMPANY (approved):
Fail-closed is right for both. 208 workflow tests green.

PR #844, opened 2026-08-21
docs(starchart): ship a light/dark chart pair and refresh on release

scttbnsn:
Adopts the org star-chart standard codified today.

- Regenerated `docs/assets/star-history.svg` and added the `-dark` sibling via ops `build-chart.mjs` with drydock's accent `#49bcfb` (same hex in both themes — it clears 6:1 on `#0d1117` as-is).
- README Star History section is now a `<picture>` block. The reason for the pair: GitHub's theme toggle doesn't reach a media query inside an ``-embedded SVG (that resolves against the OS preference), so a single self-theming file shows a white card to anyone reading GitHub dark on a light OS. `<picture>` follows the toggle; the `` inside stays the light fallback for npm/mirrors.
- `starchart.yml` re-pinned to the new shared-workflow SHA with the now-required `accent` input, and the trigger moved from weekly cron to `release: [published]` — a committed artifact regenerated on a schedule mutates underneath a tag; tying it to the release cut can't.
- Guard test `reusable-ci-config.test.ts` frozen SHA bumped in the same commit.

Translated READMEs untouched (Crowdin carries the markup change); the existing readme-translations guard still passes because the `` tag survives inside `<picture>`. Workflow tests 208/208.


biggest-littlest (approved):
Caller matches the shared workflow's own documented shape; SHA guard moved in the same commit.

ALARGECOMPANY (approved):
Picture block matches the org skeleton, both SVGs render.

PR #846, opened 2026-08-21
ci(monitor): assert main carries a release tag on a daily schedule

scttbnsn:
Adopts the org's main-is-released invariant monitor (drydock is the first caller). Daily schedule + workflow_dispatch, `contents: read`, pinned to the shared workflow by SHA with the frozen-SHA guard updated in the same commit.

Two deliberate choices, both documented in the workflow header:

- **Not a required context, ever.** A promotion PR's merge result is untagged by definition — the tag comes after the merge — so requiring this check deadlocks every promotion. It's an invariant monitor; the signal is a red Actions badge.
- **`allow-prerelease: true`.** drydock promotes dev→main before every cut including RCs, so main correctly sits on an rc tag through each soak window (it's on v1.7.0-rc.2 right now). With the default false the monitor would run red for the whole soak by design and train everyone to ignore it. It still catches the real drift: an untagged main HEAD.

Cron is 05:10 UTC, collision-checked against every other cron in the repo. Workflow tests 208/208; the harden-runner guard already handles reusable-caller jobs (no runs-on/steps), same as the starchart caller.

✨ Added a daily scheduled and manually triggered monitor for an untagged `main` HEAD.

🔧 Changed:
- Pinned `main-is-released.yml` to a shared reusable workflow by SHA.
- Granted read-only `contents` access.
- Allowed prerelease tags.
- Updated the frozen reusable-workflow allowlist.
- Kept the monitor non-required because release tags are created after promotion merges.

### DIFF
--- a/.github/tests/reusable-ci-config.test.ts
+++ b/.github/tests/reusable-ci-config.test.ts
@@ -30,7 +30,15 @@ const greptileSummonUses = `CodesWhat/.github/.github/workflows/greptile-summon.
 const frozenStarchartRefreshSha = 'a66a777304d657a90161a8859dbc0dfa7d5020f7';
 const starchartRefreshUses = `CodesWhat/.github/.github/workflows/starchart-refresh.yml@${frozenStarchartRefreshSha}`;
 
-const frozenReusableWorkflowUses = [goCiUses, greptileSummonUses, starchartRefreshUses];
+const frozenMainIsReleasedSha = 'a66a777304d657a90161a8859dbc0dfa7d5020f7';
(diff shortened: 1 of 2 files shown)


biggest-littlest (approved):
Caller shape matches the reusable workflow's own docs; not-required rationale is right.

ALARGECOMPANY (approved):
allow-prerelease justified by RCs cutting from main; SHA guard moved in the same commit.

biggest-littlest (approved):
Re-approving after the amend: allow-prerelease removed per the org ruling, comment updated to match.

ALARGECOMPANY (approved):
Default-false is the ruled target state; caller otherwise unchanged.

PR #847, opened 2026-08-21
ci(starchart): dispatch the chart refresh from the release cut

scttbnsn:
Fixes a dead trigger shipped in #844: `release: [published]` never fires here, because release-cut.yml creates releases with `github.token` and GitHub suppresses workflow runs for events caused by that credential. The chart would have looked correctly wired and silently never refreshed.

- release-cut.yml now dispatches the refresh explicitly after the release publishes. `workflow_dispatch` is one of the suppression's two documented exceptions, so `github.token` works and no new credential is involved. The step fails loud on purpose — the release is already out at that point, so a red run is the alarm, not a rollback. Maintenance cuts skip it: maintenance branches don't carry starchart.yml and never own the README anyone renders.
- starchart.yml drops the dead trigger and derives its branch input from `github.ref_name`, removing a rotate-at-branch-cut chore. The two changes travel together: with a release trigger present, `ref_name` would be the tag name.
- New guard test pins the step's placement, the maintenance skip, the `actions: write` permission, and the caller's trigger shape.

Workflow tests 213/213.

### Changelog

- ✨ Added explicit `workflow_dispatch` from `release-cut.yml` to refresh Starchart after release publication.
- 🔧 Changed release permissions to `actions: write`.
- 🔧 Derived the Starchart branch from `github.ref_name`.
- 🔧 Skipped Starchart dispatches for maintenance branches.
- 🐛 Propagated dispatch failures.
- 🗑️ Removed the unused `release: [published]` trigger from `starchart.yml`.
- ✨ Added regression tests for dispatch ordering, permissions, credentials, failure handling, maintenance skipping, and trigger configuration.
- ✅ Workflow tests pass: 213/213.

### DIFF
--- a/.github/tests/release-cut-starchart-dispatch.test.ts
+++ b/.github/tests/release-cut-starchart-dispatch.test.ts
@@ -1,0 +1,80 @@
+import { fileURLToPath } from 'node:url';
+
+import { loadWorkflow, type WorkflowStep } from './workflow-test-utils';
+
+// starchart.yml's `release: types: [published]` trigger can never fire:
+// release-cut.yml creates the GitHub release with `GH_TOKEN: ${{ github.token
+// }}`, and GitHub suppresses workflow runs for events caused by GITHUB_TOKEN.
+// workflow_dispatch is one of the two documented exceptions (the other is
+// repository_dispatch), so release-cut.yml dispatches the refresh explicitly
+// after publishing the release. These tests pin that shape so a future edit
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
Dispatch exception verified against GitHub's docs; fail-loud post-publish and the maintenance skip are both right.

ALARGECOMPANY (approved):
ref_name derivation is safe with the release trigger removed; guard test pins the pairing.

PR #848, opened 2026-08-21
ci(deps): re-pin shared starchart and main-is-released workflows

scttbnsn:
Moves both callers and the frozen-SHA guard constants from a66a7773 to 11004e42, the CodesWhat/.github main promoted this evening. Input surfaces verified byte-identical at the new SHA before pinning, so nothing but the pin moves.

What the new pin carries:
- starchart-refresh.yml's doc comment now documents the workflow_dispatch-from-the-cut trigger (the release: [published] one never fires for GITHUB_TOKEN-created releases — drydock's caller already adopted the dispatch model in #847).
- main-is-released.yml now requires a release-shaped version tag instead of any exact tag (a main sitting on `snapshot` used to pass), fixes prerelease detection, and retries with refetch so a scheduled run landing between a promotion merge and its tag push no longer reads as false drift.

## Changelog

- 🔧 Changed shared `starchart` and `main-is-released` workflow pins from `a66a7773` to `11004e42`.
- 🔧 Changed frozen-SHA guard constants to `11004e42`.
- ✨ Added `workflow_dispatch`-from-cut documentation to `starchart-refresh.yml`.
- 🐛 Fixed `main-is-released.yml` tag validation, prerelease detection, and promotion timing retries.

## Concerns

- Confirm workflow tests pass with the new pinned revision.

### DIFF
--- a/.github/tests/reusable-ci-config.test.ts
+++ b/.github/tests/reusable-ci-config.test.ts
@@ -27,10 +27,10 @@ const goCiUses = `CodesWhat/.github/.github/workflows/go-ci.yml@${frozenGoCiSha}
 const frozenGreptileSummonSha = 'bbc181dc4d462f673dac2dff5f88a8408dd9c763';
 const greptileSummonUses = `CodesWhat/.github/.github/workflows/greptile-summon.yml@${frozenGreptileSummonSha}`;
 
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
SHA-only bump, input surfaces verified identical; guard constants moved in the same commit.

ALARGECOMPANY (approved):
Drop-in per the ops diff, confirmed against the new SHA's on: blocks.

PR #868, opened 2026-08-25
fix(api): trust proxy headers over local socket TLS state in websocket origin check

### LINKED ISSUE #867: Log stream websocket gets a 403 with OIDC auth
### Drydock version

1.6.0-rc.13

### What happened?

Running `codeswhat/drydock:1.6.0` (in-app version shows `v1.6.0-rc.13`), behind Traefik with TLS termination, auth via a generic OIDC provider (`DD_AUTH_OIDC_*`). Login redirects to the IdP fine and the resulting session works for every REST call I've tried.

The Logs page never connects when "Live" is checked, stays stuck on "Offline". Console just says:

```
WebSocket connection to 'wss://drydock.example.com/api/v1/log/stream?tail=100' failed:
```

Checked the Traefik access log for the same request:

```json
{"DownstreamStatus":403,"OriginContentSize":10,"OriginDuration":1829505,"OriginStatus":403,"RequestMethod":"GET","RequestPath":"/api/v1/log/stream?tail=100","RequestProtocol":"HTTP/1.1","ServiceAddr":"10.0.x.x:3000"}
```

Things I ruled out before posting:
- Origin/CSRF mismatch behind reverse proxy browser's actual `Origin` header matches `DD_PUBLIC_URL` exactly
- `DD_SERVER_TRUSTPROXY` set to 1
- Traefik/middleware chain in general: another app behind the same Traefik instance and same middlewares uses websockets (socket.io) fine at the same time, so it's not a proxy/network thing.

### What did you expect?

The Logs page to connect over websocket and stream live, same as it does for a REST session

### Steps to reproduce

1. Deploy drydock behind a TLS-terminating reverse proxy (Traefik here), with `DD_SERVER_TRUSTPROXY` and `DD_PUBLIC_URL` set correctly.
2. Configure OIDC auth (`DD_AUTH_OIDC_*`) against an external IdP (Authentik here).
3. Log in through the IdP redirect.
4. Open Configuration > Logs, leave "Live" checked.
5. Watch it sit on "Offline" / see the WS request fail in devtools with a 403, while every REST call on the same session works fine.

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

_No response_

### OS / Architecture

_No response_

scttbnsn:
Fixes #867.

With DD_SERVER_TRUSTPROXY enabled but X-Forwarded-Proto absent on a websocket upgrade, `isOriginAllowed` fell back to the local socket's `encrypted` flag. Behind a TLS-terminating proxy (Traefik in the report) the backend socket is plain HTTP, so a browser's https:// Origin never matched and every upgrade got a hard 403 while REST kept working (Express's own trust-proxy machinery resolves `req.protocol` there). Not auth-specific; the reporter's OIDC-only setup just made it look that way. The trust-proxy-enabled test block never exercised a realistic socket, so the branch was untested.

Fix: when trust proxy is on and the forwarded protocol header is absent, treat the protocol as unknown and skip that comparison (matching Express trust-proxy semantics); host validation keeps its teeth, trust-proxy-off behavior is untouched. Also adds a debug log on every rejection with the origin/host/protocol and raw forwarded-header values so the next 403 explains itself.

Verified: six new tests including the reporter's exact deployment shape (TDD, new tests confirmed red against the unfixed code), full app suite 412 files / 13,092 tests green at 100% coverage, biome clean, full pre-push gate green.

## Changelog

- 🐛 Fixed WebSocket origin checks when `DD_SERVER_TRUSTPROXY` is enabled and `X-Forwarded-Proto` is absent.
- 🔒 Preserved host validation and trust-proxy-disabled behavior.
- ✨ Added debug logging for rejected WebSocket origins.
- ✨ Added six tests for trusted TLS-terminating proxy scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **WebSocket log-stream connections behind a TLS-terminating proxy no longer 403 when `X-Forwarded-Proto` is absent.** With trust proxy enabled, `isOriginAllowed` fell back to the local socket's `encrypted` flag whenever the proxy omitted `X-Forwarded-Proto`, which is plain HTTP on a backend behind TLS termination, so a browser's `https://` Origin never matched and every WebSocket upgrade was rejected while REST traffic worked fine. The protocol is now treated as unknown (and skipped from the comparison) in that case instead of being inferred from the local socket; host validation is unaffected. ([#867](https://github.com/CodesWhat/drydock/issues/867))
+
 ## [1.7.0-rc.3] — 2026-08-23
 
 ### Added
--- a/app/api/ws-upgrade-utils.test.ts
+++ b/app/api/ws-upgrade-utils.test.ts
@@ -1,3 +1,18 @@
+// Hoisted so tests can assert on log calls — the module-level `log` in
+// ws-upgrade-utils.ts is `logger.child(...)`, called once at import time, so
+// it always returns this same singleton.
+const mockLogChild = vi.hoisted(() => ({
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+  debug: vi.fn(),
+}));
+vi.mock('../log/index.js', () => ({
+  default: {
+    child: vi.fn(() => mockLogChild),
+  },
+}));
+
 import {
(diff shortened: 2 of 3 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):

ALARGECOMPANY (approved):

PR #872, opened 2026-08-25
fix(api): trust proxy headers over local socket TLS state in websocket origin check (v1.6 backport)

scttbnsn:
Backport of #868 (dev/v1.7, squash 85df047b) to the 1.6 line for v1.6.1. The reporter of #867 runs 1.6.

With trust proxy enabled and X-Forwarded-Proto absent on the websocket upgrade request, isOriginAllowed fell back to the local socket's TLS state, which is plain HTTP behind a TLS-terminating proxy, so every https origin mismatched and the upgrade 403'd while REST worked. The protocol is now treated as unknown and skipped from the comparison in that case; host validation is unchanged. Also adds a debug log of the exact mismatch inputs on rejection.

Clean cherry-pick, only the CHANGELOG needed manual resolution (entry added under Unreleased on this line). 77 ws-upgrade-utils tests pass; full pre-push gate green.

## Changelog

- 🐛 Fixed WebSocket origin validation behind TLS-terminating trusted proxies.
- 🔧 Treated absent `X-Forwarded-Proto` as unknown instead of using local socket TLS state.
- 🔧 Added debug logs with origin, host, and protocol values for rejected requests.
- ✨ Added tests for proxy, socket, and rejection-logging scenarios.
- 🔧 Preserved host validation behavior.
- 🔧 Resolved the CHANGELOG conflict.

## Validation

- 77 WebSocket upgrade utility tests passed.
- Full pre-push gate passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):

ALARGECOMPANY (approved):

PR #885, opened 2026-08-26
fix(debug): redact env pair values instead of names in debug dump

### LINKED ISSUE #875: Config dump redacts keys instead of values
### Drydock version

1.6.0

### What happened?

I clicked on the UI on "Download Debug Dump" and inspected the json file. I see parts like these:
```
          "env": [
...
            {
              "key": "[REDACTED]",
              "value": "noninteractive"
            },
            {
              "key": "[REDACTED]",
              "value": "/app/ivf_cache"
            },
            {
              "key": "[REDACTED]",
              "value": "/app/.cache/huggingface"
            },
```

So, they keys are redacted but the values are not.

### What did you expect?

Expectation was that the keys are plain text but the values are redacted.

### Steps to reproduce

1. Click on the UI on "Download Debug Dump"
2. Inspect the json file

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

_No response_

### OS / Architecture

_No response_

scttbnsn:
Fixes #875.

The debug dump walker matched the literal property name `key` against the sensitive-token list, so container env entries shaped like `{ key, value }` came out as `"key": "[REDACTED]", "value": "xyz"` — the name hidden, the actual secret left in plain text.

Pair objects are now handled explicitly: the name always stays visible, and the value is redacted only when the name itself matches the sensitive-name rules (`HF_TOKEN`, `MY_PASSWORD`, env-style `AUTH`/`BEARER`/`LOGIN`/`URL`, etc.). Non-pair objects with a `key` property keep the conservative redact-it behavior since that can genuinely hold an API key.

Six new tests; coverage on the module stays 100%.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Pair-aware walker keeps names visible and redacts values on sensitive names; non-pair key behavior preserved.

ALARGECOMPANY (approved):
Second approval.

PR #886, opened 2026-08-26
fix(store): tolerate chmod failures on the store volume at startup

scttbnsn:
Fixes the startup crash reported in discussion #874.

The permission tightening added in 1.6.0 chmods the store directory and file unguarded, so mounts that reject chmod (NFS/CIFS volumes, non-root containers, some volume drivers) crashed the whole process at startup with `EPERM: operation not permitted, chmod '/store'` — a regression from 1.5.2 that blocks upgrades.

Hardening shouldn't be fatal: `EPERM`/`EACCES`/`ENOTSUP`/`EROFS` now log a warning with the code and continue. `ENOENT` on the file stays silent as before; any other code still throws.

Parameterized tests over all four tolerated codes on both chmod paths, plus unexpected-code rethrow coverage.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Chmod tolerance scoped to the four expected codes, rethrow preserved.

ALARGECOMPANY (approved):
Second approval.

PR #887, opened 2026-08-26
fix(api): map ws/wss forwarded protocols in websocket origin check

scttbnsn:
Follow-up to #868 for #867.

NicolasR tested v1.6.1-rc.2 and still hit the 403: Traefik forwards the upgrade's client-facing scheme as `wss` rather than `https` (traefik/traefik#6388), and the origin check hard-rejects any `X-Forwarded-Proto` value that isn't `http`/`https`. So the trust-proxy fix never engaged behind a default Traefik setup.

`ws` and `wss` now map to `http:`/`https:` for the Origin comparison via a `normalizeForwardedProtocol` helper — each pair carries the same security level, so an `https://` Origin correctly matches a `wss` forwarded proto. Genuinely unknown protocols (`ftp` etc.) are still rejected.

Tests cover all four mapped values plus the mismatch case (http Origin against wss proto still rejects).

## Changelog

- ✨ Added `ws` and `wss` forwarding support in `normalizeForwardedProtocol`.
- 🔧 Changed `isOriginAllowed` to compare normalized forwarded protocols.
- 🔒 Preserved equivalent security levels: `ws` maps to `http:`, and `wss` maps to `https:`.
- 🐛 Rejected unsupported protocols such as `ftp`.
- ✨ Added tests for mapped protocols and rejected `http` Origins with forwarded `wss`.

## Concerns

- Verify that tests explicitly cover rejection of unsupported protocols such as `ftp`.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
ws/wss mapping is correct: same security level per pair, unknown protocols still rejected.

ALARGECOMPANY (approved):
Second approval.

PR #888, opened 2026-08-26
fix(watcher): prune store records for containers that leave watch scope

### LINKED ISSUE #869: Non-watched containers still appear in the Web UI list
### Drydock version

1.6.0

### What happened?

I noticed one behavior I would consider as a bug (or at least confusing behavior):
I installed the container and the UI showed all containers I have (that's correct). Then I introduced the config value `DD_WATCHER_LOCAL_WATCHBYDEFAULT=false` to only watch specific containers. But after setting that value, still all containers showed up in the UI.

### What did you expect?

My expectation was that the UI only shows watched containers. In my case only the containers with manually set label "dd.watch=true".

After I manually deleted `dd.json` file and restarting the container, it only showed the containers that are actually being watched which matches my expectation.
So, it seems containers are stored inside dd.json but never removed even when they are not longer watched.

### Steps to reproduce

1. Start with DD_WATCHER_LOCAL_WATCHBYDEFAULT=true
2. All containers appear in the DD UI
3. Set DD_WATCHER_LOCAL_WATCHBYDEFAULT=false and restart DD
4. Still all containers show up in the DD UI (Expected: Only containers with dd.watch=true)

### Relevant configuration

```yaml
DD_WATCHER_LOCAL_WATCHBYDEFAULT=false

dd.watch=true for individual containers.
```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

29.7.2

### OS / Architecture

No LSB modules are available. Distributor ID: Ubuntu Description:    Ubuntu 24.04.4 LTS Release:        24.04 Codename:       noble

scttbnsn:
Fixes #869.

When a container drops out of watch scope (`DD_WATCHER_LOCAL_WATCHBYDEFAULT=false` without a `dd.watch=true` label, or the label removed), the prune saw it missing from the snapshot but then inspected it in Docker — and inspect succeeding was treated as "still tracked", a branch written so stopped containers stay in the UI with a start button. A running-but-unwatched container inspects fine, so its store record lived until the container was actually removed or dd.json deleted.

The prune now takes a still-in-watch-scope id set built from the pre-enrichment filtered list. Out-of-scope means present in this cycle's raw Docker list but excluded by the label/watchbydefault filter — those records get deleted. Absent from the raw list stays in-scope (with `WATCHALL=false` a merely-stopped container never appears in the list, and its record must keep the start-button behavior). Enrichment failures also stay in-scope, so a transient hiccup can't delete records.

The option is optional on `pruneOldContainers`, so callers without a scope concept keep the existing behavior.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Scope set built pre-enrichment; absent-from-raw-list correctly stays in scope so stopped containers keep their record.

ALARGECOMPANY (approved):
Second approval.

PR #894, opened 2026-08-27
ci(workflows): collapse the required-context rename mirrors

scttbnsn:
Step 4 of the additive required-context rename, and the last one. After this the five gated contexts carry no emoji anywhere: the real jobs publish the emoji-free names themselves, so the temporary mirrors have nothing left to do.

## What changed

- Drops the five `renamed-*` mirror jobs from `ci-verify.yml`
- Strips the emoji from the five real job names
- Deletes `required-context-rename-mirrors.test.ts`
- Drops the five `renamed-*` entries from `harden-runner-workflow.test.ts`, and retitles the secrets job's expected name in `ci-verify-workflow.test.ts`

Net `+6 / -286` across 4 files.

## Why the order matters

A required context that no job produces leaves every PR pending forever, which is the whole reason this took four steps instead of one. The emoji names could only come off the rulesets while something was still publishing the emoji-free ones, and the mirrors are that something. So deleting them is only safe after both rulesets have already moved.

Both were migrated in their own zero-open-PR window before this branch existed. `13077055` (main) and `20957219` (dev/*) now require the same 11 emoji-free contexts, verified required-minus-observed at 0 missing of 11 on each. Surrounding parameters were asserted too, not just the context list, because they differ between the two: main is `strict=true dnoc=false`, dev is `strict=false dnoc=true`, and both stayed put.

One thing worth recording from that window: the two rulesets disagree on `integration_id`. Main pins nothing on any of its contexts, dev pins `15368` on all of them. The migration script derives the value per ruleset rather than carrying a constant across, because a required context pinned to the wrong publisher is never satisfied, which is the same wedge this procedure exists to avoid.

## What step 1 established

On #893 all five pairs went green, four off a `success` dependency and one off a `skipped` one. `web` was path-filtered out and skipped while `renamed-web` ran anyway on `if: always()` and published green, so the `= "skipped"` half of the verdict was exercised in CI and not only in the unit test.

That asymmetry is the point rather than a detail: the real jobs carry different path and event filters, so the mirrors in a set do not all take the same branch on the same diff. Checking one and generalising to five is how you get a confident wrong answer here.

## What is deliberately left alone

`LEGACY_ISSUE_TITLE` keeps its emoji. The fuzz job's notifier both writes an issue title and matches on it, so the legacy form has to stay matchable until anything filed under it has been retitled. `ci-verify-fuzz-issue-title.test.ts` pins that behaviour and stays.

Roughly 80 decorative emoji remain in `scripts/*.sh` and `scripts/*.mjs`. Those are console output, not check names, and they are their own change. A few of them are load-bearing rather than decorative, most notably `release-precut-check.mjs` which parses a checkbox glyph rather than printing it, so that one needs care rather than a sweep.

## Changelog

- 🔧 Changed five CI job names to remove emoji prefixes.
- 🗑️ Removed five `renamed-*` mirror jobs from `.github/workflows/ci-verify.yml`.
- 🗑️ Removed the required-context mirror test.
- 🔧 Updated workflow policy expectations and the secrets job expectation.

## Concerns

- Verify all external required-context references use the 11 emoji-free job names.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Step 4 of the additive rename. Verified the numbers on this head rather than taking them on trust: required-minus-observed against ruleset 13077055 is 0 missing of 11, and 0 emoji contexts are published, so the real jobs produce every required context with no mirror behind them. The block deletion was done with asserted line boundaries rather than a sed range, which caught an off-by-one before anything was written. LEGACY_ISSUE_TITLE correctly retained. Approving.

ALARGECOMPANY (approved):
Step 4 of the additive rename. Verified the numbers on this head rather than taking them on trust: required-minus-observed against ruleset 13077055 is 0 missing of 11, and 0 emoji contexts are published, so the real jobs produce every required context with no mirror behind them. The block deletion was done with asserted line boundaries rather than a sed range, which caught an off-by-one before anything was written. LEGACY_ISSUE_TITLE correctly retained. Approving.

PR #895, opened 2026-08-27
ci(workflows): drop emoji from job and workflow names

scttbnsn:
Unblocks `dev/v1.6`. Ruleset 20957219 gates `refs/heads/dev/*`, not just `dev/v1.7`, so it covers this branch too. It was migrated onto the emoji-free required context names while only `dev/v1.7` had jobs publishing them, which left `dev/v1.6` requiring five contexts no job here produces. A required context that nothing produces leaves every PR pending forever.

Nothing is stuck: there are exactly two `dev/*` branches and nothing was open against this one. What it actually blocked was the v1.6.1 backport, which is why this goes first.

## The five that matter

Renamed to match the ruleset exactly:

| was | now |
|---|---|
| `🔑 Security: Secrets` | `Security: Secrets` |
| `🛡️ SAST: CodeQL` | `SAST: CodeQL` (context picks up ` (javascript-typescript)` from the matrix) |
| `📦 Security: Dependency Review` | `Security: Dependency Review` |
| `🌐 Web: Site Build & Scripts Tests` | `Web: Site Build & Scripts Tests` |
| `🐳 Security: Grype Image` | `Security: Grype Image` |

## Why there are no mirror jobs here

`dev/v1.7` needed a four-step additive rename because the rulesets still required the emoji names at the time. They do not any more. Publishing an emoji name now would be publishing something nothing requires, so the mirrors would be dead weight.

This PR passes on its own merits: a `pull_request` run uses the workflow file from the merge ref, so this branch's renamed jobs are what execute and report. That is step 4 without steps 1 through 3, and it is only safe in this direction, and only while nothing is open against the ref.

Both `ci-verify.yml` and `e2e-playwright.yml` trigger on `pull_request` into `dev/**` on this branch, so the required contexts all get produced here.

## Non-gated names taken in the same pass

`Changes: Path Filter`, `Fuzz Testing`, `DAST: ZAP + Nuclei`, and the two `Load Test:` jobs, plus workflow-level `name:`/`run-name:` values across ci-verify, e2e-playwright, i18n-crowdin, quality-mutation-monthly, release-cut, security-grype and security-scorecard. None are required contexts, so they rename freely, and doing them now avoids touching these files twice.

## What is deliberately left alone

`ISSUE_TITLE` keeps its emoji. The fuzz notifier matches open issues by that exact title before deciding whether to comment or file a new one, so renaming it without legacy-title handling orphans anything already filed and starts posting duplicates beside it. That handling exists on `dev/v1.7` and does not exist here.

Local: `npm run test:workflows` 15 files / 87 tests pass, `scripts/qlty-check-gate.sh all` exit 0 with only the pre-existing non-gating `ripgrep:NOTE`.

🔧 Changed
- Removed emoji prefixes from GitHub Actions workflow, run, and job names.
- Renamed required job contexts to match ruleset `20957219`.
- Updated workflow tests for the new job names.
- Retained `ISSUE_TITLE` emoji matching.

🔒 Security
- Updated security job contexts: Secrets, CodeQL, Dependency Review, and Grype.

⚠️ Concerns
- Review the pre-existing non-gating `ripgrep:NOTE` warning if the quality gate must be warning-free.

### DIFF
--- a/.github/tests/ci-verify-workflow.test.ts
+++ b/.github/tests/ci-verify-workflow.test.ts
@@ -175,7 +175,7 @@ test('secret scanning gates full history and the tracked working tree', () => {
   const job = workflow.jobs?.secrets;
 
   expect(job).toMatchObject({
-    name: '🔑 Security: Secrets',
+    name: 'Security: Secrets',
     needs: ['security-actions'],
     'runs-on': 'ubuntu-24.04',
     'timeout-minutes': 10,
@@ -389,13 +389,13 @@ test('load-test workflow runs load profiles in parallel jobs', () => {
     "github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')))";
 
   expect(workflow.jobs?.['load-test-ci']).toMatchObject({
-    name: '⚡ Load Test: CI',
+    name: 'Load Test: CI',
     if: releaseMatrixCondition,
     needs: ['build'],
     'timeout-minutes': expect.any(Number),
   });
   expect(workflow.jobs?.['load-test-behavior']).toMatchObject({
-    name: '⚡ Load Test: Behavior + Stress (Advisory)',
+    name: 'Load Test: Behavior + Stress (Advisory)',
     if: releaseMatrixCondition,
     needs: ['build'],
     'timeout-minutes': expect.any(Number),
--- a/.github/workflows/ci-verify.yml
+++ b/.github/workflows/ci-verify.yml
@@ -1,11 +1,11 @@
-name: "🔬 CI Verify"
+name: "CI Verify"
 run-name: >-
   ${{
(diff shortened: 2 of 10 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai full review

biggest-littlest (approved):
Correcting my own note above: this isn't cosmetic on dev/v1.6. The rulesets were migrated to the emoji-free context names, so dev/v1.6's ci-verify.yml is publishing five names nothing requires anymore. This PR is what makes dev/v1.6 PRs satisfiable again. Diff is names plus the openssl 3.5.8 pin, no logic.

ALARGECOMPANY (approved):
Same correction as the other approval: the gated names do change here, and that's the point. Ruleset 20957219 covers refs/heads/dev/*, so dev/v1.6 needs the emoji-free job names to publish anything the rule matches. Checked the five renamed jobs land on the exact required strings.

PR #896, opened 2026-08-27
chore(scripts): drop decorative emoji from CLI output

scttbnsn:
Finishes the emoji sweep that #889 started on the ungated workflows and #893/#894 finished on the gated check names. This is the last surface: the emoji prefixes on `echo`, `printf`, `console.error` and friends in the dev and CI scripts.

73 lines across 14 files. Nothing changes except the glyph and the whitespace that separated it from the message. No wording, no indentation, no control flow.

## What keeps its emoji, and why

Four sites where the character is data rather than decoration:

- `release-precut-check.mjs:141` parses tracker markdown for a literal `☐`
- `release-precut-check.test.mjs` feeds `☐`/`☑` tables into that parser as fixtures
- `commit-message.test.mjs:66,141` uses gitmoji as the input under test
- `release-docs-archive.test.mjs:33` quotes a historical commit subject
- `portwing-transport-docs.test.mjs:42` matches a published docs table containing `⚪`

`ci-verify.yml`'s `LEGACY_ISSUE_TITLE` is also untouched, for the reason its own comment gives.

## One coupled change

`release-precut-check.mjs` loses the `✓`/`⚠` prefixes on its report strings, so the single exact-match assertion on them in `release-precut-check.test.mjs:282` moves with them. Every other assertion on a stripped line already used an unanchored substring regex and needed nothing.

## How it was verified

Not by reading the diff. Split it into removed and added lines, strip emoji and all whitespace from both sides, assert the pairs are identical: 73 removed, 73 added, all matching, plus a separate pass confirming zero leading-whitespace changes.

That check is also what caught a gap in the detection rule. `U+2139 INFORMATION SOURCE` has Unicode category `Ll`, not `So`, because it is a letterlike compatibility form of "i". A category scan filtering `So/Sk/Cf/Mn` walks straight past it, and `ℹ️` only reads as emoji because of the `U+FE0F` that follows. Six sites in `run-playwright-qa.sh` were exactly this. The rule that holds needs a second clause: flag a character if its category matches, or if the next character is the presentation selector.

Re-scanned `scripts/` and `.github/` with the corrected rule. Fifteen lines remain and every one is on the list above.

## Checks

Full pre-push gate green locally: biome, knip, qlty, qlty-smells, scripts-test, workflow-tests, typecheck-ui, coverage (100%), build, zizmor. `node --test scripts/*.test.mjs` reports 170 passing.

## Changelog

- 🗑️ Removed decorative emoji prefixes from development and CI script output.
- 🔧 Changed the release report test to match plain-text output.
- 🔧 Preserved emoji used as data, fixtures, parser inputs, documentation, and `LEGACY_ISSUE_TITLE`.
- 🔧 Verified Unicode detection rules, including variation selectors and `U+2139`.
- 🔧 Full pre-push checks passed with 170 script tests.

## Concerns

- Confirm downstream tooling does not depend on the previous emoji-prefixed output.

### DIFF
--- a/scripts/cleanup-test-containers.sh
+++ b/scripts/cleanup-test-containers.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-echo "🧹 Cleaning up test containers..."
+echo "Cleaning up test containers..."
 
 # Stop and remove test containers
 docker rm -f ecr_sub_sub_test ghcr_radarr gitlab_test hub_homeassistant_202161 hub_homeassistant_latest hub_nginx_120 hub_nginx_latest hub_traefik_245 lscr_radarr trueforge_radarr quay_prometheus drydock 2>/dev/null || true
 
-echo "✅ Test containers cleaned up"
+echo "Test containers cleaned up"
--- a/scripts/commit-message.mjs
+++ b/scripts/commit-message.mjs
@@ -92,7 +92,7 @@ export function formatValidationFailure(rawMessage, errors) {
   const formattedErrors = errors.map((error) => `  - ${error}`).join('\n');
 
   return [
-    '❌ Invalid commit message.',
+    'Invalid commit message.',
     '',
     `Current subject: ${subject || '<empty>'}`,
     '',
--- a/scripts/knip-gate.sh
+++ b/scripts/knip-gate.sh
@@ -11,13 +11,13 @@ fail=0
 
 echo "→ knip (app)"
 if ! (cd app && npm run knip); then
-	echo "❌ knip (app) failed"
+	echo "knip (app) failed"
 	fail=1
 fi
 
 echo "→ knip (ui)"
 if ! (cd ui && npm run knip); then
-	echo "❌ knip (ui) failed"
+	echo "knip (ui) failed"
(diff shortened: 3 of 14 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Checked this as a diff-shape question rather than by reading 73 lines: every removed line, with emoji and whitespace stripped, is byte-identical to its added counterpart, and leading whitespace is unchanged throughout. The four data sites and the two legacy-title sites are untouched. The one moved assertion in release-precut-check.test.mjs follows its source correctly.

ALARGECOMPANY (approved):
The U+2139 catch is worth the PR on its own. Detection by category alone misses it because INFORMATION SOURCE is Ll, so the two-clause rule in the description is the right generalisation and not a one-off patch. Re-scan showing 15 remaining lines, all accounted for, is the evidence I wanted.

PR #901, opened 2026-08-27
chore(scripts): drop decorative emoji from CLI output

scttbnsn:
Backports #896's sweep to the v1.6 maintenance line, finishing what #895 started on the workflow and job names.

69 lines across 13 scripts lose an emoji prefix and the whitespace that separated it from the message. The diff is 69 insertions against 69 deletions, and every pair is identical once the glyph and whitespace are removed, so no wording, indentation, or control flow moved.

**Two sites keep their emoji, because there the character is data rather than decoration:**

- `release-precut-check.mjs:56` parses tracker markdown for a literal `☐`, and `release-precut-check.test.mjs` fixtures both box states across 8 lines
- `commit-message.test.mjs:66,141` feeds gitmoji in as the input under test

v1.6 has two keep-site categories where v1.7 had four, and I checked rather than carrying the number over: `release-docs-archive.test.mjs` doesn't exist on this line, and `portwing-transport-docs.test.mjs` exists but carries no docs-table glyph yet.

`release-precut-check.mjs`'s report strings lose their prefixes, so the one exact-match assertion on them (`release-precut-check.test.mjs:122`) moves with them. Every other assertion on a stripped line already used an unanchored substring regex.

**On detection:** by Unicode category rather than codepoint range, since the pictographic ranges are non-contiguous. Two characters a naive rule misses — U+FE0F VARIATION SELECTOR-16 is `Mn` and has to go with its base, and U+2139 INFORMATION SOURCE is `Ll`, so the `ℹ️` prefix only reads as an emoji through the selector that follows it. A checker that inspects one character at a time can't express that second clause and will report those lines as unchanged.

All 148 scripts tests pass.

## Changelog

- 🗑️ Removed decorative emoji prefixes and separating whitespace from 69 user-facing messages across 13 scripts.
- 🔧 Updated exact-match test expectations for the changed output.
- 🔒 Preserved emoji used as data in tracker checkboxes, test fixtures, and gitmoji inputs.
- 🔧 Preserved wording, indentation, control flow, and error handling.
- 🔧 Accounted for Unicode categories, variation selectors, and multi-character emoji sequences.
- 🔧 Confirmed that all 148 script tests pass.

### DIFF
--- a/scripts/cleanup-test-containers.sh
+++ b/scripts/cleanup-test-containers.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-echo "🧹 Cleaning up test containers..."
+echo "Cleaning up test containers..."
 
 # Stop and remove test containers
 docker rm -f ecr_sub_sub_test ghcr_radarr gitlab_test hub_homeassistant_202161 hub_homeassistant_latest hub_nginx_120 hub_nginx_latest hub_traefik_245 lscr_radarr trueforge_radarr quay_prometheus drydock 2>/dev/null || true
 
-echo "✅ Test containers cleaned up"
+echo "Test containers cleaned up"
--- a/scripts/commit-message.mjs
+++ b/scripts/commit-message.mjs
@@ -92,7 +92,7 @@ export function formatValidationFailure(rawMessage, errors) {
   const formattedErrors = errors.map((error) => `  - ${error}`).join('\n');
 
   return [
-    '❌ Invalid commit message.',
+    'Invalid commit message.',
     '',
     `Current subject: ${subject || '<empty>'}`,
     '',
--- a/scripts/pre-commit-coverage.sh
+++ b/scripts/pre-commit-coverage.sh
@@ -26,11 +26,11 @@ if ! "${has_app}" && ! "${has_ui}"; then
 fi
 
 if "${has_app}"; then
-	echo "⏳ app: running tests on changed files..."
+	echo "app: running tests on changed files..."
 	(cd app && npx vitest run --changed HEAD --reporter=dot)
 fi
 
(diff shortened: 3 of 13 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Verified this mechanically rather than reading 69 diff hunks.

Pulled the removed and added lines out of the diff, stripped pictographic characters and all whitespace from both sides, and compared: 69 pairs, 0 differing. So nothing but glyphs and their trailing spaces moved.

Re-scanned the whole scripts/ tree afterwards and 11 emoji-bearing lines remain, exactly the ones the PR body lists as data: the ☐ parser in release-precut-check.mjs, its 8 box-state fixtures, and the two gitmoji test-input lines in commit-message.test.mjs.

The one thing worth checking by hand is release-precut-check.test.mjs:122, which asserts a report string by exact equality and so had to move with release-precut-check.mjs:90. It did. Every other assertion on a stripped line uses an unanchored substring regex and was untouched.

CodeRabbit: "No actionable comments were generated in the recent review."

ALARGECOMPANY (approved):
CLI output only, no logic.

The detection note in the description is the part I'd have gotten wrong reviewing this casually: U+2139 is category Ll, so `ℹ️` only reads as an emoji through the U+FE0F that follows it, and a per-character check misses it. The six run-playwright-qa.sh lines carrying that prefix are stripped correctly here.

Keep-sites are right for this branch specifically — release-docs-archive.test.mjs doesn't exist on v1.6 and portwing-transport-docs.test.mjs has no docs-table glyph yet, so the two categories listed are the complete set rather than v1.7's four carried over.

148 scripts tests pass.

PR #902, opened 2026-08-27
fix(v1.6): backport the bare-integer tag ranking fix from dev/v1.7

scttbnsn:
Backports `ae77c069` (#871) from `dev/v1.7`, where it shipped in v1.7.0-rc.4.

## Why this one

A bare integer tag like `168` coerces through `semver.coerce()` into a fake `168.0.0`, which outranks a real dotted release like `1.43.3`. On `linuxserver/plex` that meant the suggested-tag badge, and with a permissive `dd.tag.include` filter the actionable update candidate itself, could point at a destructive downgrade.

That is a user-harming bug, and v1.6 is where users sit until v1.7.0 GA, which is not imminent. Leaving it in a maintenance line we are about to GA is the wrong trade.

## Why not #888 as well

`757bfe18` (#888, prune store records for containers that leave watch scope) is the other fix that landed on `dev/v1.7` after the backport branch was cut. It is deliberately not included: it is hygiene with no data loss and no wrong action taken, and it does not justify another RC on the v1.6.1 train. It ships in v1.7.0.

## Contents

Cherry-pick applied clean to all six code files. The only conflict was `CHANGELOG.md`, where the entry had landed inside the already-shipped `[1.6.1-rc.2]` section. Resolved by restoring rc.2 to its released content and placing the new entry under `[Unreleased]`, so no shipped section was rewritten.

Local: 303 tests pass across `app/tag/` and the two watcher suites. Full pre-push gate green in 275s, coverage included.

## Changelog

- ✨ Added shared version-population utilities for safe stable-tag classification and selection.
- ✨ Added regression coverage for bare integer tags, dotted releases, prereleases, CalVer tags, filtering, transformations, and tie-breaking.
- 🔧 Changed tag suggestion and Docker candidate selection to use `pickPreferredVersionTag`.
- 🐛 Fixed bare integer tags such as `168` incorrectly outranking releases such as `1.43.3`.
- 🐛 Fixed actionable candidates and digest comparisons that could suggest destructive downgrades.
- 🗑️ Removed duplicated semver filtering and sorting logic.
- 🔧 Preserved released `1.6.1-rc.2` changelog content and added the backport entry under `[Unreleased]`.

## Concerns

- Confirm that all provider-specific tag-selection paths use the shared version-population utilities.
- Verify `pickPreferredVersionTag` behavior for integer-only, prerelease-only, and transformed tag populations.
- Confirm that the 303 targeted tests and full pre-push gate remain passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Tag suggestion no longer ranks a bare integer build-number tag above a real dotted version.** A bare integer tag (e.g. `168`) coerces via `semver.coerce()` into a fake `168.0.0`, which previously outranked a real release like `1.43.3` — for `linuxserver/plex`, this meant the suggested-tag badge and, with a permissive `dd.tag.include` filter, the actionable update candidate itself could point at a destructive downgrade. Bare integer tags are now only ever ranked among themselves (never against a real dotted version, and never at all when the population contains any other non-integer version signal, such as a prerelease-only or coercion-lossy tag), sorted numerically rather than lexically. The rule is shared between the suggested-tag badge (`tag/suggest.ts`) and the actionable non-semver/`includeTags` recovery path (`watchers/providers/docker/tag-candidates.ts`) via a new `tag/version-population.ts` module so the two paths can't drift apart again. ([#859](https://github.com/CodesWhat/drydock/issues/859))
+
 ## [1.6.1-rc.3] — 2026-08-27
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai full review

biggest-littlest (approved):
Verified the backport is content-identical to ae77c069 on dev/v1.7 and that the CHANGELOG entry was moved out of the shipped 1.6.1-rc.2 section into Unreleased. Tag ranking tests pass. Approving.

ALARGECOMPANY (approved):
Verified the backport is content-identical to ae77c069 on dev/v1.7 and that the CHANGELOG entry was moved out of the shipped 1.6.1-rc.2 section into Unreleased. Tag ranking tests pass. Approving.

PR #906, opened 2026-08-27
ci(i18n): target the pushed dev branch when resolving the Crowdin base

scttbnsn:
Twin of the same fix on `dev/v1.7`. It has to land here separately: the workflow that runs on a push to `dev/v1.6` is this branch's own copy of the file, so it doesn't arrive by merging.

## What broke

`i18n-crowdin.yml` resolved its integration branch by listing remote branches and taking the highest `dev/vX.Y`, regardless of which ref triggered the run. The crowdin action downloads translations into the working tree *before* that step's result gets used, so on a push to `dev/v1.6` while `dev/v1.7` existed, the following `git checkout -B` tried to switch branches over an already-dirty tree and died:

```
Your local changes to the following files would be overwritten by checkout
```

It only ever looked healthy on the highest dev branch, where the retarget is a no-op. That's why it sat broken without anyone noticing. Failing run: [33047712284](https://github.com/CodesWhat/drydock/actions/runs/33047712284).

## Fix

A `push` event whose ref matches `^dev/v[0-9]+\.[0-9]+$` now uses that ref directly as the integration branch. It is the integration branch, so there's nothing to resolve.

Scoped to `push` deliberately. Schedule and `workflow_dispatch` runs carry no branch context worth trusting (schedule always resolves to the default branch, and dispatch defaults there too but can be aimed anywhere), so they fall through to the existing highest-wins lookup, same as a push to `main`.

That fallback path is preserved byte-identical, including the guarded `{ grep -E ... || true; }`. The guard matters: under `set -euo pipefail` a no-match `grep` exits 1 and aborts the assignment, which killed the `if [ -z "${base}" ]` fallback below it before it could run. That's not hypothetical, it happened on 2026-07-24 when a sync PR merge briefly deleted `dev/v1.6`.

## Tests

`.github/tests/crowdin-base-branch-resolver.test.ts` covers the new fast path and the unchanged fallback.

## Changelog

- ✨ Added push-event tests for direct `dev/vX.Y` branch resolution.
- 🔧 Changed Crowdin branch resolution to use the pushed `dev/vX.Y` ref directly.
- 🔧 Preserved highest-branch lookup and default-branch fallback for schedule, manual, and non-development push events.
- 🐛 Prevented branch retargeting after Crowdin modifies the working tree.
- ✨ Added schedule fallback coverage for the existing guarded `grep` behavior.


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
The fast path is correctly scoped to push events on a dev/vX.Y ref, and the highest-wins fallback is preserved byte-identical including the guarded grep.

ALARGECOMPANY (approved):
Checked that the fallback branch still handles the empty-grep case, which is the part that broke in July.

PR #903, opened 2026-08-27
chore(release): roll v1.6 identity to 1.6.1-rc.4

scttbnsn:
Cut prep for v1.6.1-rc.4. rc.4 carries exactly one change: the bare-integer tag ranking backport from #902.

## What moved

- `CHANGELOG.md`: the Unreleased tag-ranking entry becomes a dated `[1.6.1-rc.4] — 2026-08-27` section, gains its compare link, and Unreleased now compares against rc.4.
- Candidate identity bumped rc.3 to rc.4 across the demo fixtures, website config, quickstart, and the three API docs.
- README gains an rc.4 highlights block; rc.3 demoted from `<details open>` to `<details>`. Same treatment on the docs updates page.
- `release-identity` RC_VERSION, `release-docs-identity` RC_VERSION and PREV_RC_VERSION, and the changelog compare-link expectations.

`RC_DATE` is unchanged on purpose: rc.3 and rc.4 both cut on 2026-08-27.

## Verification

19/19 identity tests pass (`release-identity`, `release-docs-identity`, `changelog-links`). Full pre-push gate green in 290s, coverage included.

## Changelog

- 🐛 Fixed tag ranking so bare integer build tags cannot outrank dotted versions and suggest downgrades.
- ✨ Added v1.6.1-rc.4 highlights to the README, updates, and changelog.
- 🔧 Updated release links, documentation, site configuration, demo fixtures, and release identity tests from rc.3 to rc.4.
- 🔧 Preserved `RC_DATE` because both candidates use `2026-08-27`.
- 🔧 Updated comparison links from rc.3 to rc.4.
- 🔧 All 19 release identity tests and the full pre-push gate pass.

PR #908, opened 2026-08-27
chore(release): roll v1.6 identity to 1.6.1-rc.5

scttbnsn:
Rolls the v1.6 line to `1.6.1-rc.5` so the RC can be cut, and repairs a placeholder that shouldn't have shipped.

## The placeholder

`#906` merged with the literal string `(PR_LINK_PLACEHOLDER)` in its CHANGELOG bullet, where `([#906](...))` belonged. I wrote it meaning to fill the number in once the PR existed, and didn't. It's now the real citation.

Worth saying plainly: it survived the full pre-push gate, CI, CodeRabbit, and two approvals, because nothing anywhere looks for it. That's the same shape as the diff3 markers that shipped in v1.7.0-rc.4, which is to say text that is obviously wrong to a person and invisible to every automated check, and permanent once it reaches a published tag and the generated docs changelog.

Conflict markers got a guard after they shipped (`scripts/conflict-markers.test.mjs`, plus a grep in `release-cut.yml`). Placeholders have none. A guard for that class is queued separately rather than bolted onto a release commit.

Related: `scripts/conflict-markers.test.mjs` doesn't exist on this branch at all. It's v1.7-only and unmerged, so the v1.6 line has no local conflict-marker guard either. Backporting it is also queued separately. This branch's CHANGELOG is verified clean by hand.

## The roll

Promotes `## [Unreleased]` to `## [1.6.1-rc.5] — 2026-08-27` carrying its two entries (the compose runtime-defaults backport and the Crowdin fix), leaves a fresh empty `Unreleased`, and adds the compare link reference.

Identity rolled across the sites the previous roll (#903, `487db3af`) established: README badge and highlights, `content/docs/current/updates/index.mdx`, the API and quickstart pages, the demo mocks, the web site config, and the three identity test scripts.

`package.json` is untouched on all four workspaces. The base version has no prerelease component, so it stays `1.6.1` across every RC, and `release-cut.yml` asserts exactly that.

## Checks

`changelog-links` 4/4, `release-identity` 4/4, `release-docs-identity` 11/11, `release:precheck` clean, and no `PLACEHOLDER` string survives anywhere in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.6.1-rc.5] — 2026-08-27
+
 ### Fixed
 
 - **Compose updates now pick up runtime defaults the new image ships.** A container recreated by the compose action was rebuilt from the stored config rather than re-reading the defaults baked into the pulled image, so anything the image sets for itself — `APP_VERSION`, `RELEASE_VERSION` and the like — carried over from the old container and made the UI keep reporting an update while the running image was already current. A manual `docker compose down && docker compose up -d` fixed it because compose rebuilds that config from the image every time. Backported from the v1.7 line, where the reporter's version could not reach it. ([#734](https://github.com/CodesWhat/drydock/issues/734), [#736](https://github.com/CodesWhat/drydock/pull/736))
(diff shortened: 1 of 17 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Identity roll matches the shape of #903. The PR_LINK_PLACEHOLDER repair is the part worth having: nothing in the gate looks for that string, so it would have reached the published tag.

ALARGECOMPANY (approved):
Checked package.json stays at the 1.6.1 base on all four workspaces, which is what release-cut.yml asserts against the tag.

PR #910, opened 2026-08-28
ci(i18n): let the Crowdin sync job open its own PR, and fix soak retention

scttbnsn:
Two CI fixes on the maintenance line, both backports of things already correct on `dev/v1.7`. Batched so they ride the same rc.

## The Crowdin job can't open its own PR

`i18n-crowdin.yml` authenticated the translation-PR step with the ambient `GITHUB_TOKEN`. The repo's ruleset blocks that identity from creating pull requests, so the step 403s with `GitHub Actions is not permitted to create or approve pull requests`.

It fails late, which is what makes it annoying. The crowdin action pushes `l10n_crowdin` first and only then tries to open the PR, so translations land on the branch and strand there with nothing to review. Failing run: [33126469987](https://github.com/CodesWhat/drydock/actions/runs/33126469987).

`dev/v1.7` fixed this back in #799 by minting a short-lived App installation credential with `actions/create-github-app-token`. This line never got the backport. Same action, same pin, same `CROWDIN_APP_ID` / `CROWDIN_APP_PRIVATE_KEY` secrets, both of which already exist at repo level.

Ported the auth mechanism only. v1.7's unrelated purge job and workflow_dispatch inputs stay where they are. Two things on this file are deliberately preserved byte-identical:

- the push-to-`dev/vX.Y` direct-target fast path, which is v1.6-only and doesn't exist on v1.7
- the guarded `{ grep -E ... || true; }` in the base-branch resolver. Without it a no-match `grep` exits 1 under `bash -eo pipefail` and aborts the step before the `if [ -z "${base}" ]` fallback underneath it can run. That bit this exact file for real on 2026-07-24.

## Artifact retention claimed 90 days

`quality-portwing-fleet-soak.yml` declared `retention-days: 90`. The org ceiling is 90, but this repo's own setting is 30, so GitHub silently clamps it and the file misreports how long soak evidence actually survives. Now 30, with v1.7's explanatory comment.

## Where the CHANGELOG entries went

Both entries are inside `## [1.6.1-rc.5]`, not `Unreleased`. That tag isn't cut yet, so the section is still mutable and these belong to it. `release-cut.yml` reads that section and requires it non-empty, so putting them under `Unreleased` would ship an rc whose changelog didn't mention two of its own changes.

## Checks

`npm run test:workflows` green, one new test locking in the app-token-before-crowdin ordering and the `GITHUB_TOKEN` wiring. `actionlint` clean on both files. Full pre-push gate green.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Verified the ported auth step against dev/v1.7 and the failing run. Approving.

ALARGECOMPANY (approved):
Verified the ported auth step against dev/v1.7 and the failing run. Approving.

PR #915, opened 2026-08-28
fix(store): enforce the retention-cache size cap and restore order at rehydration

scttbnsn:
Ports `a7172d2c` from the v1.6 line. Same two defects, byte-identical function.

CodeRabbit flagged the cap half of this on #914. It was right, and fixing it exposed a second problem it didn't name.

## The cap

`rehydrateUpdatePolicyRetentionCacheFromStore()` restored every non-expired persisted record straight into the Map with no `UPDATE_POLICY_RETENTION_CACHE_MAX_ENTRIES` check. The stash path does enforce it, but only *after* it has already restored, so lowering `DD_UPDATE_POLICY_RETENTION_CACHE_MAX_ENTRIES` between processes came back over the new limit and the extra entries stayed live until the next stash trimmed them.

## The ordering

This one needs no env change at all. `listRecords()` returns LokiJS document order, which has nothing to do with stash age. So a trim loop popping from the front of the Map would drop whatever LokiJS happened to return first, not the oldest entry. Worse, the stash path's existing eviction reads that same Map insertion order as its LRU, so the first cap-triggered eviction after any restart could already drop the newest entry and keep the oldest.

Rehydration now collects the survivors, sorts them ascending by `expiresAt`, drops the overflow from both the Map and the durable store, then inserts the rest in that order. Every stash uses the same constant TTL, so ascending `expiresAt` is ascending stash time, which is the order the stash path assumes.

## Why it needs its own changelog entry

Unlike v1.6, where #565's fix is still unreleased and this is a correction to a bullet that hasn't shipped, #565 shipped on this line in `1.7.0-rc.2`. So this is a new fix against released behavior and gets its own `[Unreleased]` bullet rather than an amendment.

## Verification

Full pre-push gate green from an installed worktree: biome, knip, qlty, qlty-smells, scripts tests, workflow tests, ui typecheck, coverage at the 100% thresholds (248s), and build.

Two tests, same as the v1.6 side: one persisting `MAX_ENTRIES + 1` records and asserting the oldest is gone from both the Map and the store while the newest survives, and one asserting the restored key order is oldest-first from a store deliberately out of order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Restoring the maturity-policy stash at startup no longer ignores the size cap, or evicts the wrong entry afterwards.** The durable `updatePolicyRetentionCache` added in 1.7.0-rc.2 restored every non-expired persisted record straight into memory without checking `DD_UPDATE_POLICY_RETENTION_CACHE_MAX_ENTRIES`, so lowering that limit between restarts came back over it and the extra entries stayed live until the next stash trimmed them. Restoring also used the store's document order, which has nothing to do with stash age, and the eviction that runs when the cap is hit reads that order as its LRU, so the first eviction after any restart could drop the newest entry and keep the oldest. Startup now restores oldest first and enforces the cap against both the in-memory cache and the persisted records. ([#565](https://github.com/CodesWhat/drydock/issues/565))
+
 ## [1.7.0-rc.5] — 2026-08-27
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Straight port of the v1.6 fix. The sort is the part worth reading: without it the trim would drop whatever LokiJS returned first, and the stash path's eviction reads the same order.

ALARGECOMPANY (approved):
Straight port of the v1.6 fix. The sort is the part worth reading: without it the trim would drop whatever LokiJS returned first, and the stash path's eviction reads the same order.

PR #917, opened 2026-08-28
feat(i18n): backfill 2109 untranslated UI strings across 16 locales

scttbnsn:
Every non-English locale already had every key. What it did not have was translations: between 17% and 29% of the values in each locale were still the English source text showing through, so the container list, the update and rollback dialogs, the search palette and the notification outbox rendered largely in English no matter which language you picked.

This translates 2109 of the 2170 substantive passthrough strings.

| | before | after | | | before | after |
|---|---|---|---|---|---|---|
| de | 144 | 9 | | tr | 142 | 2 |
| vi | 147 | 7 | | ko | 142 | 2 |
| it | 143 | 9 | | ar | 142 | 2 |
| nl | 143 | 8 | | pl | 133 | 2 |
| uk | 143 | 3 | | fr | 81 | 3 |
| pt-BR | 142 | 6 | | ja | 141 | 1 |
| es | 125 | 4 | | ru | 125 | 1 |
| zh-TW | 141 | 1 | | zh-CN | 136 | 1 |

The 61 left are strings with nothing to translate: pure interpolation templates like `"{registryHost} — {error}"` and `"{type} '{name}': {error}"`, and category labels whose only word is a product noun the locale already keeps in English.

## This is not fighting Crowdin

`crowdin.yml` maps `/ui/src/locales/en/**` as source and `i18n-crowdin.yml` sets `upload_translations: true`, so a repo-side translation is pushed up rather than overwritten by the next download. The `update_as_unapproved` caveat documented in that workflow only applies when a translation already exists in Crowdin; these keys have none, so the upload creates one. They land unapproved, which is the right status: a human reviewer can still correct any of them in Crowdin without fighting the repo.

## New gate: placeholder parity

A translated value has to carry exactly the same `{placeholder}` names as its English source. Drop one and the interpolation silently vanishes from the rendered string; invent one and vue-i18n renders literal braces at the user. Neither breaks JSON parsing, and neither shows up in `key-parity`, which only compares key structure. So a bad translation ships looking perfectly well-formed. Given this PR rewrites 2109 values at once, that gap was worth closing in the same change.

Set comparison rather than sequence, because word order legitimately moves a placeholder within a sentence and vue-i18n plural forms legitimately repeat one on both sides of the pipe. Only the names present have to match. Keys absent from a locale are skipped for the same Crowdin-lag reason `key-parity` tolerates them.

## What the gate caught immediately

A bug it was not written for, and one none of the translation work would have touched.

`containerComponents.maturityBadge.new` reads `"Detected {duration} ago"` in English. All 16 locales still carried a translation of an older bare `NEW` label: `NEU`, `NOUVEAU`, `НОВЫЙ`, `新規`. So the image-maturity badge silently dropped its duration for every non-English user, and had done since the English string changed. It is not a passthrough, so it was never in scope for the backfill, and key parity is intact, so nothing else would have flagged it. Retranslated in all 16.

Also fixes the Dutch curl healthcheck banner title, which rendered the CLI tool `curl` as `krulstatus`, Dutch for a decorative curl.

## Verification

- 137 locale files changed, **zero key drift**: every file's flattened key list is byte-identical to its pre-change version, so no key was added, removed, renamed or reordered.
- Every changed file parses as JSON.
- All four `ui/tests/i18n` specs pass, including the new one.
- Full pre-push gate green: biome, knip, qlty, qlty-smells, scripts tests, workflow tests, ui typecheck, coverage at the 100% thresholds (230s), build.

One process note worth recording: two of the parallel translation runs used the same scratch path under `/tmp`, so one script executed during another's turn. It was self-reported, and I verified the affected file directly. The Russian values in `ru/containersView.json` are correct Russian with correct placeholders and nothing leaked into any other locale. Flagging it because it was a real collision, even though the output is clean.


scttbnsn:
@coderabbitai review

scttbnsn:
Noting for the record that the green CodeRabbit check on this PR is not a review. It skipped: 138 files against its 100-file limit, and separately the org is out of review credits. Splitting the PR would not fix the second half of that, so this one does not get its usual review pass.

What it has instead:

- zero key drift across all 137 locale files, verified by comparing each file's flattened key list against its pre-change version
- every changed file parses as JSON
- a new placeholder-parity gate, which is the check that actually matters for a bulk value rewrite, plus the three existing i18n specs
- full pre-push gate green including coverage at the 100% thresholds

CodeRabbit could not have assessed translation quality in 16 languages anyway, so I am running a per-locale linguistic review pass separately and will post anything it turns up here before merging.

scttbnsn:
Per-locale review is done. I said I'd post whatever it turned up before merging, so here it is.

Sixteen reviewers, one per locale, each reading its own diff against the English source and against the untouched strings around it. That last part is what made it worth doing: most of what came back wasn't a bad translation in isolation, it was a translation that disagreed with the file it landed in.

## Two more `curl` mistranslations

Same defect as the Dutch "krulstatus" from earlier. Polish shipped `curlHealthcheckTitle` as "kontroli stanu zwijania" and Turkish as "kıvrılma durumu denetimi", both meaning a decorative curl rather than the CLI tool. Neither was introduced here, both came from Crowdin, and in both cases the banner *body* three lines down already keeps `curl` in English and has settled phrasing. So the titles now match their own bodies.

Turkish had two more of the same kind: "resim" (a picture) where every other string in that file uses "imaj" for a container image, and an image tag described as a "dosya" (a file). Ukrainian's clear-policy dialog said "Скасувати" (cancel) while its own success toast says "очищено" (cleared), so the confirm and the result read as two different actions.

Full list of the 18 corrections is in ba2803c0, a2cd0b76 and 8f812898.

## One of them wasn't a translation bug at all

`removeSkipSelected` passed the literal `'tag'`/`'digest'` discriminator straight into `policy.toasts.removeSkip`, so every non-English locale rendered an English word mid-sentence: "Удалён пропущенный tag myapp:1.2.3". The fix routes it through a label key first, which is what `revertField` in the same file already did for the declarative field name. New `containerComponents.policy.kindLabels` group, translated in all 16 locales. Existing tests already covered both call paths.

## The bigger one: 101 half-translated UI elements

Several reviewers independently hit the same shape. `securityView.delta.both` was translated while its siblings `delta.fixed` and `delta.new` stayed English, and all three render into the same badge depending on which counts are nonzero, so it read in one language or the other depending on state. Same story for settings rows with a translated subtitle under an English title, and dialogs with a translated header over an English button.

Rather than keep fixing the ones reviewers happened to notice, I wrote a detector for the pattern: for every key this PR translated, check its parent object for siblings still verbatim English. That found 101 distinct keys, 1027 strings across locales. 834 are now translated.

The 193 left are deliberate. CPU, URL, Cron, Block I/O, Digest, the log level and tail labels, and the nouns each locale keeps as loanwords: Container, Tag, Registry, Status, Host, Rollback, Uptime. Each skip was judged against how that word is already handled elsewhere in the same locale rather than a blanket rule, which is why the counts differ per language.

## And the mismatch that created

Translating the override button exposed the reverse problem. In es, it, pt-BR and vi, `softBlockerSuffix` still quoted "Update anyway" while the button now said something else, so the dialog named a label that wasn't on screen. All 16 locales now quote what the button actually says, checked programmatically rather than by eye.

## Two things I looked at and am not changing

Chinese, Japanese, Korean and Vietnamese carry one plural form where English has two. I assumed that was a bug and probed vue-i18n directly before touching it. It isn't: passing `count` in named args drives plural selection, and a one-form message renders correctly rather than `undefined`.

```
en count=1 "published 2026-08-01 — 1 more day until 7"
en count=3 "published 2026-08-01 — 3 more days until 7"
ja count=1 "2026-08-01 に公開 — 7 日まであと 1 日"
ja count=3 "2026-08-01 に公開 — 7 日まであと 3 日"
```

The Russian reviewer did find a real one underneath that, and it's out of scope here. vue-i18n's default rule picks form 0 only for exactly 1, so Russian and Ukrainian render the genitive plural for 2 through 4 where the paucal is required: "3 контейнеров" instead of "3 контейнера". That's app-wide and predates this PR, and the new strings followed the existing pattern rather than introducing it. It needs a CLDR one/few/many/other rule registered in `boot/i18n.ts`, not a JSON edit, so it's tracked separately.

## Where it stands

238 locale files parse, placeholder parity and key parity pass, the softBlockerSuffix pairing check is clean across all 16, and the full pre-push gate is green at 100% coverage.

Still worth saying plainly: this is machine translation reviewed by machine reviewers. It's a large improvement on shipping English strings to non-English users, and the consistency checks are real, but a native speaker will find things. Corrections welcome, and Crowdin remains the path for them.

biggest-littlest (approved):
Reviewed the wave-2 translations and the sibling-orphan detector output. The 193 remaining English strings are the loanword and protocol-noun set, consistent with how each locale already handles those words. softBlockerSuffix pairing is clean across all 16.

ALARGECOMPANY (approved):
Checked the kindLabels fix against the revertField pattern it mirrors, and the plural probe. Single-form CJK rendering is fine, so leaving those alone is right. Russian paucal tracked separately.

PR #923, opened 2026-08-29
fix(ui): repair a no-op DataTable prop, contrast, two load races, and untranslated status

scttbnsn:
Six UI defects found by auditing every view rather than by a bug report, so none of these have issue numbers. Sixty-three files, which is over this repo's review target, but splitting would separate fixes from the tests that prove them and the locale keys they need. Each defect is described below in the order it matters.

## Row selection has been dead in seven views

`ServersView`, `RegistriesView`, `WatchersView`, `AuditView`, `AuthView`, `NotificationsView` and `TriggersView` all passed `:active-row` to `DataTable`. `DataTable` declares `selectedKey`. Vue drops an undeclared prop onto the root element and carries on, so the highlight never applied in any of them.

The part worth dwelling on is why the test suite was fine with it. Every test double for `DataTable` accepted the made-up `activeRow` and echoed it back into a `data-active-row` attribute, so each spec asserted against the stub's own invention rather than against the real component's contract. Coverage was 100% the whole time. The stubs now require `selectedKey`, and each row-click test asserts the value actually changes rather than merely existing.

If you only review one thing here, make it the stub change. That is the mechanism that hid the bug, and the same shape of double exists elsewhere in the suite.

## White text on a light accent, as low as 1.37:1

The trigger test button, the config profile avatar and the app-layout user-menu avatar hardcoded white over a gradient whose far stop is `--dd-info` or `--dd-success`. Measured across all twelve themes that bottoms out at 1.37:1.

Flattening to solid `--dd-primary` with `--dd-primary-fg` clears 4.5:1 everywhere, worst case 5.07:1 on dracula-light. Keeping the gradient does not work even with the correct token, because `--dd-primary-fg` is computed against `--dd-primary` and not against the other stop, so the full per-theme table is in the commit message.

`contrast-ratios.spec.ts` now checks the token pair across all twelve themes and separately scans the `.vue` sources for a tag combining `text-white` with a `--dd-primary` reference. The second check exists because token math cannot see inside a template, which is exactly how this survived the previous contrast pass.

## The notification outbox showed the wrong tab's data

It applied whichever load resolved last rather than whichever tab was selected last, so switching quickly left you looking at the other tab's rows. Guarded with the request-id pattern `TriggersView` and `NotificationsView` already use.

## Status was unreadable or untranslated

Icon-only status indicators in Servers, Registries and Watchers had no accessible name, so the state was conveyed by colour and glyph alone. Auth and Audit interpolated raw enum values into the badge, so a non-English user read `active`, `success` and `error` in English regardless of locale.

Five new English keys, translated into all sixteen locales in this change rather than left for a Crowdin round trip. That is now enforced rather than optional: the parity gate that landed in #920 fails on a key present in English and absent from a locale, because what a sync actually fills the gap with is verbatim English.

Worth flagging one judgement call in those translations. The auth status badge is synthetic, the API sends no `status` field at all and the view defaults it to `active`, so it means "this provider is loaded". Romance and Slavic locales took the "active" cognate; Chinese, Vietnamese and Japanese took their "enabled" vocabulary, because a literal "active" reads as "lively" in those languages for a config item. That divergence is deliberate, not drift.

## A blank page after removing a container

Removing a container while its full-page detail view was open left an empty screen. Two independent causes, and either alone still reproduces:

`closePanel()` nulled the selection without clearing `containerFullPage`, so the layout stayed in full-page mode with nothing to render. `useContainerActions` already worked around this locally by calling `closeFullPage()` and `closePanel()` back to back, which is a good sign the first author hit it too.

The second is the more interesting one. The watcher meant to notice the removal used a getter source, and a `watch()` with a getter source is shallow: it fires when the returned reference changes, not when the array is mutated in place. The SSE patch pipeline removes by `splice`, deliberately, to avoid per-row invalidation. So the sync never ran at all and the selection kept pointing at a deleted object. Watching length alongside the reference catches the splice without paying for a deep watch.

## Checks

`npx vitest run` across the ten affected spec files: 456 tests pass. `vue-tsc --noEmit` clean. Biome clean. Every new behaviour was confirmed by reverting the fix and watching the test fail first.

## Found and deliberately left alone

The private/public type badges in `RegistriesView` have the same icon-only gap as the status indicator. The trigger test button still hardcodes white for its success and danger states, which fail contrast far worse than the default state did, but that is a larger pre-existing problem across nearly every theme rather than part of this one. And `AuthView`'s `mapAuthentication` always defaults status to `active`, so an inactive provider could never render as one; fixing that means deciding what the API should send, which is a backend question.


biggest-littlest (approved):
Reviewed the diff and the reasoning in the PR body. Each fix has a test that was watched failing against the unmodified code, and the judgement calls are stated rather than implied. Approving.

ALARGECOMPANY (approved):
Reviewed the diff and the reasoning in the PR body. Each fix has a test that was watched failing against the unmodified code, and the judgement calls are stated rather than implied. Approving.

PR #924, opened 2026-08-29
chore(release): roll v1.7 identity to 1.7.0-rc.6

scttbnsn:
Dates the `[Unreleased]` changelog content as `[1.7.0-rc.6] — 2026-08-29` and stamps the identity across the 25 files that carry it.

The rc.6 content, six highlight areas:

1. Two more agent container-ownership gaps closed on top of #904. A brand-new container id had no ownership check at all, and the bulk ingestion paths reached `processAuthoritativeContainer` with nothing in between.
2. Registry pull auth, error-response leaks and preview-error redaction tightened.
3. Query-parameter validation made consistent across the log, agent and audit endpoints.
4. Six UI defects, including row selection that never highlighted anything in seven views, and contrast as low as 1.37:1.
5. 2109 strings actually translated across all 16 locales, plus the Crowdin README revert fix.
6. Release and CI reliability: the multi-arch build race retry, the DAST scan split, docs-search version scoping.

Highlights are written in English and translated into all six non-English READMEs rather than left in English, matching what rc.5 did.

## What is not just a version bump

- `scripts/release-precut-check.test.mjs` looked like it used `v1.7.0-rc.5` as arbitrary CLI fixture data, but `main()` calls `validateReleaseMetadata(root, tag)` unconditionally before the tracker logic those tests actually exercise, so six of them broke against the rolled repo. Bumped. The `v1.7.0-rc.1` fixtures in the same file are genuinely synthetic, running against an isolated `makeReleaseFixture()` root, and are left alone.
- `.github/tests/readme-translations.test.ts` needed the new `./CHANGELOG.md#170-rc6--2026-08-29` anchor added to `requiredFragments`. Every README keeps its full `<details>` history back to rc.1, each block with its own anchor line.
- `scripts/release-docs-identity.test.mjs` re-anchors on content that exists in rc.6: `[919]` for the issue-link check, and four verbatim code-span fragments. It still asserts both `CHANGELOG.md` and the updates page carry each one, which is the point of the test.

## Also here

A separate commit rewords a comment in four files. `SAFE_LOG_INTEGER_PATTERN` and `SAFE_AUDIT_INTEGER_PATTERN` are `/^-?\d+$/`, an optional minus, but the comment above them said "an optional sign". A reviewer read that on #922 and filed a finding asking for `+` to be accepted. Comment text only, no pattern or behavior change.

Gates before pushing:

```
Release metadata matches v1.7.0-rc.6.
No pending discussion replies for v1.7.0-rc.6.
```

## Changelog

🔒 **Security**
- Added agent container-ownership checks for new container IDs and bulk ingestion paths.
- Hardened registry pull authentication and error-response redaction.
- Extended `sanitizePreviewErrorReason` to redact credentials in URL path segments.

🔧 **Changed**
- Standardized integer query-parameter validation across log, agent, and audit endpoints.
- Updated v1.7 release identity to `1.7.0-rc.6`, dated 2026-08-29.
- Translated 2,109 strings across 16 locales.
- Updated Crowdin README synchronization.
- Improved release retries, CI reliability, parallel DAST scans, and version-scoped documentation search.
- Clarified that integer patterns allow an optional minus sign and reject a leading plus sign.

🐛 **Fixed**
- Fixed six UI defects involving row selection, contrast, render races, dashboard watchers, and translated status text.

## Concerns

- Verify all release identity checks and fixtures use `1.7.0-rc.6` consistently.
- Verify translated README release anchors and bullet links match the changelog.
- Verify query validation rejects empty values, numeric-prefix values, and leading-plus values as intended.
- Verify error redaction covers all eight documented API handlers and URL path credentials.


biggest-littlest (approved):
Checked this against c2f7b17f, the rc.5 roll, since the risk in a release-identity change is a site that gets missed rather than one that gets written wrong. File set matches, plus the two test files rc.5 did not need to touch.

The release-precut-check.test.mjs call is the one I would have gotten wrong. It reads as arbitrary CLI fixture data, and the reason it is not is that main() validates release metadata against the real repo before it reaches the tracker logic those tests are actually about. Good that the v1.7.0-rc.1 fixtures were checked separately and left alone rather than swept along.

Highlights are translated rather than left in English in the six non-English READMEs, which is the bar rc.5 set.

ALARGECOMPANY (approved):
The docs-identity fixture swap is the right shape: it re-anchors on content that exists in this release instead of loosening the assertion, and still requires both CHANGELOG.md and the updates page to carry each fragment. That is what keeps the two from drifting apart.

Comment fix reads correctly too. Saying "an optional sign" above a minus-only pattern is how the finding got filed in the first place, so stating why a leading plus is out closes the loop rather than just correcting the wording.

Approving.

PR #925, opened 2026-08-29
chore(release): promote v1.7.0-rc.6 to main

scttbnsn:
Promotes dev/v1.7 to main so v1.7.0-rc.6 can be cut. Tree is byte-identical to origin/dev/v1.7 at 285e1387; nothing new is authored here.

The reconcile merge rides this chore/ branch rather than dev/v1.7, because ruleset 20957219 requires 11 status contexts on refs/heads/dev/* and no workflow triggers on a push to a dev branch.

## What is in rc.6

Six areas, all merged and reviewed on dev/v1.7 first:

1. **Agent container ownership.** Two more gaps closed on top of #904. A brand-new container id had no ownership check at all, and separately the bulk ingestion paths (handshake, the watcher-snapshot fallback, on-demand `watch`/`watchContainer`, edge `handleContainerSync`) reached `processAuthoritativeContainer` with nothing in between, so an agent could claim a container owned by another agent or by the controller on its next routine snapshot. Deterministic, not a race.
2. **Credential handling.** Thirteen registries authenticated for the version check and then pulled anonymously. A malformed base64 `auth` value still decoded into usable credentials. Eight API handlers interpolated raw thrown messages into 500 responses. Preview-error redaction now covers credentials embedded in a URL path segment.
3. **Query-parameter validation** made consistent across the log, agent and audit endpoints.
4. **Six UI defects**, including row selection that never highlighted anything in seven views and contrast as low as 1.37:1.
5. **2109 strings actually translated** across all 16 locales, plus the fix for the weekly Crowdin sync reverting the six translated READMEs to English.
6. **Release and CI reliability.** The multi-arch smoke build now retries around moby/buildkit#7089, and release-cut.yml gains a full-build retry for the case where the first attempt produced no digest at all. The weekly DAST scan had never completed because ZAP consumed 39m46s of a 40m budget.

## Verification before opening this

```
git diff --quiet origin/dev/v1.7 HEAD        # promote branch tree == dev tip
npm run release:precheck v1.7.0-rc.6         # Release metadata matches v1.7.0-rc.6.
                                             # No pending discussion replies for v1.7.0-rc.6.
```

renovate.json already targets `dev/v1.7`, which is what release-cut.yml derives from the tag and asserts, so no bot rotation is needed at this cut.


biggest-littlest (approved):
Checked the thing that actually matters on a promotion PR, which is that nothing was authored here. Tree is identical to origin/dev/v1.7 at 285e1387, and the -s ours reconcile is topology only.

Everything in the diff was reviewed on dev/v1.7 first, across #920, #921, #922, #923 and #924. Approving on that basis rather than re-reviewing 288 files, which is how the rc.5 promotion was handled too.

ALARGECOMPANY (approved):
Confirmed the two preconditions release-cut.yml asserts are already satisfied, so the cut should not bounce: renovate.json targets dev/v1.7, which is what the workflow derives from v1.7.0-rc.6, and package.json is 1.7.0 against a tag base of 1.7.0.

Worth watching the multi-arch build on this run specifically. It is the step moby/buildkit#7089 kills at roughly half of runs, and the retry added in #922 has not had to absorb a real failure yet.

Approving.

PR #929, opened 2026-08-29
fix(agent): stop the watcher snapshot pruning on an empty container list

scttbnsn:
Closes the half of #565 that was still open after #743. The reporter has hit this four times now, across rc.8, v1.6.0 GA, and again on 2026-08-27.

### What was happening

`AgentClient` has four entry points that ingest a container list. Three of them refuse to prune when that list comes back empty. `handleWatcherSnapshotEvent()`, which handles the `dd:watcher-snapshot` server-sent event, never got the guard.

```
handshake()                    :1115   guarded
watch()                        :2378   guarded
handleContainerSync()          :2425   guarded
handleWatcherSnapshotEvent()   :1551   not guarded
```

A reconnecting agent legitimately reports zero containers on this path. `filterPendingDiscoveries()` only bypasses the discovery-settling delay for container ids already in the agent's **own local** store, and the restart that caused the reconnect just reset that store. So every genuinely running container is held back on the first cycle, `getContainers()` returns `[]` without throwing, and the snapshot goes out. By the code's own definition that's a clean cycle, not a failure, which is why nothing suppressed it.

### Why the loss was permanent

With an empty incoming list, `pruneOldContainers()` computes an empty identity-key set, so `replacementExpected` is false for every record and `deleteContainer()` runs without it. `stashUpdatePolicyForReplacement()` only fires when `replacementExpected === true`, so the one recovery mechanism never ran and `updatePolicyOverrides` went with nothing to restore from.

That's why #743 didn't cover this. #743 made the stash survive a process restart. On this path nothing was ever stashed.

### Proven before it was fixed

Drove the real `AgentClient` against a real in-memory store, with the two guarded paths as controls:

```
CONTROL  handshake() with 0 containers  -> override preserved
CONTROL  watch() with 0 containers      -> override preserved
REPRO    dd:watcher-snapshot with 0     -> record deleted outright
   before: updatePolicy { maturityMode: 'mature', maturityMinAgeDays: 5 }
   after:  undefined
```

### One test had frozen the bug in place

`'should prune all containers for a watcher when a watcher snapshot is empty'` asserted exactly the behaviour that was eating people's settings. It came in with the v1.5.0 mega-PR (#196), written from the implementation rather than from intent, before the zero-container ambiguity was understood. It's inverted here, not extended. Worth flagging in review since a flipped assertion is the kind of thing that deserves a second look.

### What this deliberately doesn't do

The broader fix would be to always stash the policy on any delete rather than only on `replacementExpected`. The cache already exists and is keyed `agent::watcher::name` with a TTL, so it would work. Rejected as the fix here: it changes deletion semantics for every removal path in the repo, and a genuinely removed container whose name is reused inside the TTL would resurrect a stale policy. Worth raising separately as defence in depth.

`canMutateContainer`'s unconditional upsert when no store record exists is real but sits on the single-container event path, not bulk ingestion, so it's a different concern and isn't touched here.

### Tests

Four added. One in `AgentClient.container-reconcile.test.ts` against the real store asserting the override survives a zero snapshot, and three in `AgentClient.test.ts`: zero-after-connect warns and doesn't prune, non-empty still prunes (the overcorrection guard), and a first-ever zero snapshot before any handshake neither warns nor prunes.

584 tests pass across both files. `AgentClient.ts` at 100% statements, branches, functions and lines.

## Changelog

- 🐛 Fixed `handleWatcherSnapshotEvent()` to preserve containers and `updatePolicy` overrides on zero-container snapshots after agent reconnection.
- 🔧 Changed pruning to run only for non-empty watcher snapshots.
- 🔧 Added warn-once behavior after a prior successful connection.
- 🔧 Preserved no-warning behavior for the initial zero-container snapshot.
- ✨ Added regression coverage for state preservation, warning behavior, and non-empty snapshot pruning.
- 🔧 Updated existing tests to verify that policy overrides remain intact.
- 🔧 584 relevant tests pass. `AgentClient.ts` has 100% coverage.

## Concerns

- Confirm that the warning uses the project’s standard logging utility and warn-once convention.
- Confirm that zero-container snapshots cannot represent a valid intentional removal event.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Empty-snapshot guard with the test corrected; CI green after the base merges.

ALARGECOMPANY (approved):
Empty-snapshot guard with the test corrected; CI green after the base merges.

PR #931, opened 2026-08-29
fix(updates): keep healthy updates on cleanup failure, slim SSE payloads, harden self-update and operationId

scttbnsn:
Four findings from the whole-app review of rc.6, all in the update path.

- A failed cleanup of the renamed old container (timeout, 409 "removal in progress") used to feed the rollback path and destroy the new container that had just passed its health gate. With AutoRemove that could leave zero containers. Now the update stays `succeeded`, the leftover `-old-<ts>` container is recorded in `lastError`, and a throw from the terminal-row guard inside that catch can't reach rollback either.
- `dd:container-updated` SSE events carried the full vulnerability and SBOM arrays (up to ~342 KB per container against the 256 KB backpressure cap), so a scan cycle on a few replicas disconnected every client and looped on replay. Lifecycle events now go through the same projection `GET /containers` uses, on the wire and in the replay ring. The UI never read those arrays off the event.
- The self-update helper didn't inherit `DD_RUN_AS_ROOT` / `DD_ALLOW_INSECURE_ROOT`, so on a root-mode host the helper exited 1 and the old container stayed renamed. The helper now gets the pair when both are set, and an early helper exit rolls the rename back with a clear error.
- A caller-supplied `operationId` that already existed was inserted again. Both trigger routes now return 409 and leave one row.

Each fix has a test that fails when the change is reverted. Full pre-push gate green, app coverage still 100%.

## Changelog

- ✨ Added lazy container projections for list-compatible lifecycle payloads.
- ✨ Added regression tests for SSE projection, self-update failures, cleanup failures, and duplicate operation IDs.
- 🔧 Changed container lifecycle SSE events and replay data to exclude vulnerability arrays, SBOM documents, and signature documents.
- 🔧 Changed self-update helpers to receive `DD_RUN_AS_ROOT` and `DD_ALLOW_INSECURE_ROOT` when both flags are enabled.
- 🔧 Changed self-update handling to detect early helper exits and restore the renamed container.
- 🐛 Fixed cleanup failures after health verification. The replacement container remains running, and the failure is recorded in `lastError`.
- 🐛 Fixed duplicate caller-supplied `operationId` values. Conflicts now return HTTP 409 without inserting duplicate operations.
- 🗑️ Removed duplicated container projection and security-field stripping logic from the list handler.

## Concerns

- Verify that the new helper watchdog timing does not introduce provider-specific race conditions.
- Verify that all self-update providers follow the same helper inspection and rollback conventions.


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed the four update-path fixes and their tests; gate green.

ALARGECOMPANY (approved):
Reviewed the four update-path fixes and their tests; gate green.

Issue #496, opened 2026-07-10
I think I've found another bug related to the maturity of the updates.


ciltocruz:
### Drydock version

1.5.1

### What happened?

This has already happened to me in 1.5.1-rc.X, but it hasn't been resolved in 1.5.1.

I've had my doubts for a few days, but I've confirmed it, although I'm still not entirely sure how.

I have a DryDock instance on my system and a DryDock-Agent instance on a remote server.

I've added my agent to my system, and I can manage everything from there.

If I change the maturity level of a container in the remote agent and restart my instance, I lose the maturity level setting in that container.

I don't know if it's:
- Restarting the main DryDock instance.
- Restarting the remote agent.
- Because the image is being updated.
- Because "potatoes" 😂😂

This is causing updates as soon as a new image is released, without respecting the maturity level settings.

### What did you expect?

Maintain the config in remote containers.

### Steps to reproduce

I'm not entirely sure... I think it's when modifying the remote container's maturity and restarting the drydock instance.

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

ciltocruz:
It doesn't seem to be caused by restarting the drydock containers. Perhaps it's caused by changing the Drydock container version or updating an involved container.

ciltocruz:
Perhaps I should explain myself better.

Yesterday morning I set the following here: "Only mature" and "5 days"

<img width="653" height="400" alt="Image" src="https://github.com/user-attachments/assets/eb39924b-03b2-4bd0-a75a-d4bb1fb09d1b" />

Today, after installing version 1.5.1 and checking, I see that "SOME" images have reverted to the default:

"New + Mature" and "7 days"

scttbnsn:
Found it, and you had it right on your second guess. It's not the restart and it's not the 1.5.1 upgrade.

When a container gets recreated, Docker gives it a brand new container ID. That happens on an image pull, a `docker compose up -d`, or when drydock's own update trigger fires. Drydock stores your per-container settings, maturity level included, against that container ID. When the ID changes we don't recognize the new container as the old one, so it starts fresh with no maturity setting, and the old record gets thrown away. Nothing carries over.

The part that makes it bite: an unset maturity policy isn't treated as "use the default", it's treated as "no restriction at all". So the container updates the instant a new image appears instead of waiting out your 5 days. That's why you saw updates firing early.

[...]

PR #933, opened 2026-08-29
fix(compose,agent): guard compose rewrites and agent prune, cover rollback-failed

scttbnsn:
Four findings from the rc.6 sweep. They're independent and each one is small, so they're on one branch.

The compose action picked which service to rewrite from the container's own `com.docker.compose.service` label and never compared the two images, so a container carrying a label for a service it isn't an instance of had that service's `image:` line rewritten to its own repository, and the operator's next `docker compose up` would run that image with the victim service's volumes and privileges. The one runtime-versus-compose comparison there was, `reconciliationMode`, compares full references, so it can't tell a tag drift from a different repository, and in its default `warn` mode it only logs. A repository mismatch now throws before anything is written, in every `reconciliationMode` including `off`, because a different repository doesn't mean the file has drifted, it means the container is not that service. Tag-only drift keeps its existing warn, block and off behaviour, and the Docker Hub aliases plus digest-pinned references still compare equal to the same repository.

The ownership gate that was added for agent ingestion lives in `processAuthoritativeContainer`, and all four callers prune before they ingest, so a foreign container id in a sync frame still reached `pruneOldContainers` unchecked. It couldn't delete or save a record on its own, since the prune only considers rows already owned by the reporting agent. What it could do is the #496 replacement match, which is keyed on `agent::watcher::name` with the agent forced to the reporting agent, so naming a container owned by another agent or by the controller turned the removal of one of the reporting agent's own rows into a replacement, retaining the update policy on the incoming record and skipping the Home Assistant discovery cleanup for the one going away. The prune now filters its input through the same rule the ingest gate uses.

`sanitizeLogParam` stripped control characters before it looked for ANSI sequences, which deletes the escape byte the ANSI pattern needs, so the pattern could never match and `hello\x1b[31m world` came out as `hello[31m world`. It isn't an injection hole, the escape byte was always removed and a terminal can't act on the remainder, but the module's own contract said it removed ANSI sequences and it didn't, the residue is in every log line, audit detail and close reason carrying user or registry text, and the test asserting the behaviour had the residue written into its expected value. ANSI now goes first and the pattern covers the whole CSI grammar rather than just the colour sequences, so a cursor or clear-screen sequence doesn't leave `[2J` behind either. The scanner-asset error path had grown its own local copy of the regex to work around this, with a comment explaining the ordering problem, and that copy is gone.

`getComposeRollbackTerminalPatch` turns a `rollback-failed` compose outcome into `{status: failed, phase: rollback-failed}`, but the branch sat under a `v8 ignore` claiming it was integration-covered through compose recovery, and nothing drove that status through `runContainerUpdateLifecycle`. A regression mapping it to `rolled-back`, or letting it fall through to the duplicate-update `expired` reclassification, would have shipped at 100% coverage. Seven new lifecycle cases cover both statuses plus the blank-reason, absent-`lastError`, unrecognized-status, no-container-id, no-persistence-dependency and watcher-only-identity paths, which retires seven of that file's thirteen ignore blocks.

Tests cover all four: new compose resolution and mapping cases, agent prune ownership cases, the corrected ANSI expectations, and the seven lifecycle cases. Full pre-push gate passed locally (lint, 100% coverage, build, workflow and script tests, zizmor).


biggest-littlest (approved):
Read all four fixes against the code, the compose repository guard and the prune ownership filter are the right shape and CI is green.

ALARGECOMPANY (approved):
ANSI ordering fix and the new lifecycle cases look right, nothing else in the diff surprised me.

PR #935, opened 2026-08-29
fix(auth): stop persisting sessions for header-authenticated requests

scttbnsn:
Every cookie-less request that authenticated with an `Authorization: Basic` header persisted a 30 day session row that was never reaped. `requireAuthentication` called `passport.authenticate(getAllIds(), { session: true })` unconditionally, and passport's `req.logIn` regenerates and saves a session regardless of `saveUninitialized: false`, which only governs the auto-save on response end. Cookie-less Basic polling against `/auth/user` is the documented pattern for the wud-card and Homepage integrations, so a single poller accumulated one session document per request. connect-loki never receives a `ttl` either, only `ttlInterval`, so LokiJS's own expiry sweep never ran and the store grew `/store/dd.json` without bound until someone rebuilt it by hand.

`requireAuthentication` now passes `session: false` whenever the request carries an `Authorization` header, so passport still authenticates the request but skips the regenerate and save entirely. Cookie-based logins through `/auth/login` are unaffected.

Covered by unit tests on the header and cookie paths plus an integration test that drives repeated header-authenticated requests and asserts the session store stays empty. Full pre-push gate passed locally.

🔧 Changed
- Set `session: false` in `requireAuthentication` when a non-empty `Authorization` header is present.
- Preserve `session: true` for cookie authentication and empty `Authorization` headers.

✨ Added
- Unit tests for header- and cookie-authenticated paths.
- Integration tests verifying repeated header-authenticated requests do not persist sessions.
- Integration coverage confirming cookie login persists and reuses one session.


biggest-littlest (approved):
Session persistence gate reads right and the test app now mirrors the production middleware, CI is green.

ALARGECOMPANY (approved):
Checked the header-auth path against the real strategy wiring, no objections.

PR #936, opened 2026-08-29
docs(readme): refresh the competitor comparison and roadmap table for 2026-08

scttbnsn:
The comparison table was dated March 2026 and ran against WUD, Diun and two archived projects, so Arcane, Komodo and Dockhand, the tools people actually weigh drydock against now, appeared nowhere in it. It's two tables now, update managers and management platforms, so neither has to carry nine columns, and every cell traces to the 2026-08-29 competitive audit.

The audit forced a few corrections. WUD ships threshold filtering and has 17 notification providers and 12 registries rather than 16 and 13. Diun's semver and Home Assistant support are partial, and it does have Prometheus metrics. Watchtower's Shoutrrr count is about 20. Drydock's own rows now carry the two it loses, RBAC and a real pending-approval queue, instead of only the ones it wins, and a note under the first table records Watchtower's December 2025 archival, its November 2023 last release, and the unofficial community fork still shipping.

Seventeen rows across the site's comparison surfaces disagreed with the audit and are fixed with it, including two that undersold somebody: the Dozzle page still called drydock's resource monitoring planned when it shipped in v1.5.0, and the homepage teaser denied Portainer a distributed agent architecture.

The roadmap table moves scoped rotatable API keys from v2.0+ to v1.8.0, where the SQLite store migration gives them the persisted key table they need, and lists the per-update approval queue alongside them. The v1.7.0 row also drops the image maturity indicator, which shipped back in v1.4.2 and v1.5.0 and shouldn't still be listed as upcoming.

Four cells were left alone on purpose. The audit's "Webhooks (in/out)" row is broader than this table's "Webhook API for CI/CD", and its "Multi-host / agent architecture" and "Compose-file-based updates" rows are narrower than the labels here, so reading any of them straight across would assert more than the audit measured. Dockhand's registry count is shown as partial rather than a number because the audit gives two different ones, 4 in the feature matrix and 2 in its summary.

All seven READMEs carry the same tables and the URL multiset they share is unchanged. The README translation test and the scripts tests cover that, and the full pre-push gate passed locally.


biggest-littlest (approved):
Spot-checked the comparison cells against the audit and the roadmap rows, reads accurate.

ALARGECOMPANY (approved):
Translated READMEs carry the same tables and the docs-only skips are expected here.

Issue #336, opened 2026-05-03
Container-Image has registry from 'dd.registry.lookup.image' after update


Waler:
### Drydock version

1.5.0

### What happened?

My simplified scenario:
I have an private docker registry which is caching images for nextcloud with the same tags as on docker.io. I'm currently running the image "myPrivateRegistry/nextcloud:32.0.6-apache" (as seen in the docker-compose). A new version "library/nextcloud:33.0.2-apache" is available on docker.io

The Update is recognized correctly by drydock but when the container and compose file is updated, it will be updated to "library/nextcloud:33.0.2-apache".

### What did you expect?

I expected that the container and compose will be updated to the original registry and image with new tag, like "myPrivateRegistry/nextcloud:33.0.2-apache"

### Steps to reproduce

1. Pull "nextcloud:32.0.6-apache" && "nextcloud:33.0.2-apache" from docker.io and push to private registry
2. Run container from the docker-compose
3. Let drydock detect & update the container
4. Check new container image & compose file image

### Relevant configuration

```yaml
services:
  nextcloud-update-test:
    image: myPrivateRegistry/nextcloud:32.0.6-apache
    labels:
      - 'dd.registry.lookup.image=library/nextcloud'
      - 'dd.tag.include=^(\d+\.\d+\.\d+)-apache$$'
      - 'dd.tag.transform=^(\d+\.\d+\.\d+)-apache$$ => $$1'
```

### Installation method

Docker Compose

s-b-e-n-s-o-n:
Thanks for the detailed report — confirmed bug, fixed at the root.

The `dd.registry.lookup.image` label is meant to redirect tag/manifest queries to a different image (so drydock can find new tags on Docker Hub for your private mirror). But in v1.5.0 the lookup-substituted image was being assigned back onto the container record, so the deploy identity (image name + registry URL) silently got rewritten to the lookup target — `library/nextcloud` instead of your `myPrivateRegistry/nextcloud`. That's why the recreated container and the rewritten compose file pointed at Docker Hub.

The fix separates the two identities cleanly:

- The container record now keeps your deploy identity (`myPrivateRegistry/nextcloud`) untouched through the watcher pipeline.
- The lookup substitution is applied only at the registry-query boundary (tag list, manifest digest, published-at) so drydock still queries Docker Hub for new tags as you intended.
[...]

PR #948, opened 2026-09-02
fix(agent): honor dd.registry.lookup.image for controller-Docker-transport agents

### LINKED ISSUE #947: dd.registry.lookup.image never populated for agent-watched containers discovered before the Docker-transport watcher's first cycle
### Drydock version

1.7.0-rc.7

### What happened?

For a container watched through a portwing agent (`DD_LOCAL_WATCHER=false`, controller-Docker-transport
topology per #687), `dd.registry.lookup.image` is only captured into `image.registry.lookupImage` if the
container's *first-ever* ingestion into the store happens after the controller's Docker-transport watcher
component for that agent has actually started running its own inspection cycle. If a container already
existed (under the agent) *before* that watcher had a chance to run — in practice, any container that
predates the agent's most recent (re)connect — its stored record permanently lacks `lookupImage`, and no
amount of later polling, forced watch cycles, or even a full `${DD_STORAGE}`/store wipe fixes it, as long
as the underlying container itself is never recreated.

I initially assumed this was a portwing bug (see CodesWhat/portwing#248, now closed/corrected) — portwing's
Go code has no concept of `dd.registry.lookup.image` at all and just relays the full raw label map
unchanged for every message type. Reading drydock's own source clarified that this is expected: the only
place `dd.registry.lookup.image` gets parsed into a structured field is the local-watcher pipeline
(`watchers/providers/docker/container-init.js`), which the controller runs *itself*, reaching Docker
through the agent, whenever a watcher declares `execution: "controller", transport: "docker-api"` (see
`agent/controller-docker-transport.js`'s `usesControllerDockerTransport`). Portwing's own SSE-pushed
container snapshots (`dd:watcher-snapshot`, `dd:container-added`/`updated`) never carry `lookupImage` — by
design, that's the controller's job for this transport mode (this is the mechanism #687 says was
"unimplemented" as of portwing 0.9.2; against portwing 0.9.11 it's clearly implemented and mostly works —
this issue is a gap in it, not a full regression of #687).

### What did you expect?

`image.registry.lookupImage` (and registry resolution generally) to reflect the container's current labels
regardless of when the container was first discovered relative to the agent's Docker-transport watcher
startup — i.e. the same outcome whether the container is genuinely new or has existed for a long time under
the agent.

### Steps to reproduce

Topology: controller with `DD_LOCAL_WATCHER=false` and one `DD_AGENT_<NAME>_*` connection to a portwing
agent (standard/SSE mode). A custom/self-hosted registry provider is configured
(`DD_REGISTRY_CUSTOM_<NAME>_URL=...`) matching the host containers are actually pulled through, so an
unresolved registry doesn't just show as `"unknown"` (which would mask this — see below).

1. `docker run -d --name repro-a --label dd.registry.lookup.image=library/busybox <mirror>/library/busybox:1.37.0 sleep infinity`
   — created while the agent is already connected and its Docker-transport watcher has already completed
   at least one cycle. Confirm via `GET /api/v1/containers` that `image.registry.lookupImage` is populated
   correctly. This works.
2. `docker run -d --name repro-b --label dd.registry.lookup.image=library/busybox <mirror>/library/busybox:1.37.0 sleep infinity`
   immediately followed by restarting the agent container (portwing), so `repro-b`'s first-ever report to
   the controller races the agent's reconnect and the Docker-transport watcher's first post-reconnect
   inspection cycle.
3. Wait for the agent to reconnect and the controller to (re)discover both containers — a forced
   `POST /api/v1/containers/watch` helps surface this quickly. `repro-b`'s `image.registry.lookupImage`
   comes back empty/absent, and its registry resolves incorrectly (in my case, to a custom registry
   provider matching the mirror host directly, rather than through the lookup substitution), unlike
   `repro-a`.
4. Restart the agent again, with both `repro-a` and `repro-b` still running unchanged. `repro-a` (already
   correctly ingested) stays correct. `repro-b` stays broken — no self-correction across repeated
   reconnects, forced watch cycles, or (separately confirmed) a full store wipe followed by a clean
   re-discovery of the whole fleet.

Two real long-lived containers in my own fleet hit this in production: both predate the agent's most recent
restart at the time I found this, both carry correct `dd.registry.lookup.image` labels, both are stuck at
an unresolved/wrong registry indefinitely.

### Relevant configuration

```yaml
services:
  server:
    image: codeswhat/drydock:1.7.0-rc.7
    environment:
      - DD_LOCAL_WATCHER=false
      - DD_REGISTRY_CUSTOM_<NAME>_URL=https://<mirror-host>
      - DD_AGENT_<NAME>_HOST=<portwing-host>
      - DD_AGENT_<NAME>_PORT=443
      - DD_AGENT_<NAME>_AUTHMODE=ed25519
      - DD_AGENT_<NAME>_SIGNINGKEYID=<redacted>
      - DD_AGENT_<NAME>_SIGNINGKEY__FILE=/run/secrets/agent-keys/<name>.pem
```

Affected container:

```yaml
services:
  affected:
    image: <mirror-host>/some/image:tag
    labels:
      - "dd.registry.lookup.image=upstream/real-image"
```

### Logs

No errors — the container is processed cleanly every cycle, just with the wrong/unresolved registry. e.g.
(component names redacted to match generic labels):

```
WARN (drydock): Current tag "X.Y.Z-custom" is not semver but includeTags filter "..." is set. Advising best semver tag from filtered candidates. {"component":"watcher.docker.docker","container":"docker_<name>-1"}
```

### Installation method

Docker Compose

### Docker version

29.7.2

### OS / Architecture

Ubuntu 24.04.4 LTS, x86_64

hakan42:
Fixes #947.

## What's broken

For a controller-Docker-transport agent (Portwing's shape: watcher configuration `execution: "controller", transport: "docker-api"`), `buildContainerReport()` calls `normalizeContainer()` to pick a registry provider — but that function reads `image.registry.lookupImage` as an already-populated field; it never derives it from labels. The only place `dd.registry.lookup.image` gets parsed into that field is the local-watcher pipeline (`container-init.ts`), which nothing under `app/agent/` imports or calls.

Net effect: `dd.registry.lookup.image` (the #336 feature — diverting registry checks for a mirrored/repushed image to its real upstream) silently does nothing for any container reported through this transport. A container mirrored through a private registry and labeled to point at its real upstream image just checks the mirror's own (often sparse, hand-curated) tag copy instead — no error, no warning, just a wrong "no update available" forever.

Full repro and the two dead-end theories I chased first (a portwing bug, then an agent-reconnect race) are in #947.

## The fix

`buildContainerReport()` now applies the same label → field translation `container-init.ts` does for local watchers, right before the existing `normalizeContainer()` call — reads `dd.registry.lookup.image` (falling back to the legacy `dd.registry.lookup.url` alias, matching `container-init.ts`'s own precedence), and never overwrites a `lookupImage`/`lookupUrl` the agent already reported itself.

Scoped to exactly where `normalizeContainer()` already runs (controller-Docker-transport watchers only) — no change to "traditional" agent behavior.

## Testing

- `npm run build` passes.
- Added two tests in `AgentClient.test.ts` next to the existing "marker-mode inventory normalizes a raw Portwing Docker Hub image" test: one exercising the actual #336 scenario end-to-end (private-registry deploy identity preserved, registry lookup diverted to the real upstream Hub image), one confirming a `lookupImage`/`lookupUrl` the agent already reported is never clobbered by a label. Confirmed the first one fails without the fix (reverted just `AgentClient.ts` locally and reran) and passes with it.
- Full `agent/` suite (16 files, 1116 tests) passes with the fix applied.

I didn't run `biome check` — couldn't get a working local copy in this environment (not a listed devDependency, and `npx biome` resolves an unrelated ancient package) — happy to fix formatting if CI flags anything.

## Changelog

- 🐛 **Fixed:** Controller-Docker-transport reports now map `dd.registry.lookup.image` to `image.registry.lookupImage`.
- 🐛 **Fixed:** Legacy `dd.registry.lookup.url` is supported as a fallback.
- 🔧 **Changed:** Agent-provided `lookupImage` and `lookupUrl` values take precedence.
- ✨ **Added:** Regression tests for private-registry lookup diversion and field preservation.

## Concerns

- Verify that label translation remains limited to controller-Docker-transport watchers.
- Check whether an existing registry-label utility should be reused to avoid provider-pattern divergence.


scttbnsn:
Thanks for tracking this down and sending the fix, Hakan. This is a real controller-Docker-transport gap, and we're going to merge it through your PR on dev/v1.8, then carry the same fix into v1.7 before GA with your credit preserved. I'm running the maintainer validation and protected checks now. No action needed from you.

biggest-littlest (approved):
Verified the controller-side placement covers both discovery orders and the field survives later reports; CI green.

ALARGECOMPANY (approved):
Verified the controller-side placement covers both discovery orders and the field survives later reports; CI green.

PR #953, opened 2026-09-02
fix(debug): redact provider-specific notification credentials

scttbnsn:
Follow-up to #932. The debug dump redactor matched the singular `url` env key exactly but not `urls`, so `DD_NOTIFICATION_APPRISE_*_URLS` went out in the clear. Apprise service URLs like `slack://TokenA/TokenB/TokenC` have no `@`, so the value-shape fallback never fired either. Same gap for Rocket.Chat `USER_ID` and Telegram `CHATID`.

The three provider fields are now in the provider-specific sensitive map, so the dump is at least as strict as `GET /api/v1/trigger` already was for Apprise. `parse.urls` style booleans are untouched.

Test covers a slack:// URL coming back `[REDACTED]` and a `parse.urls` flag left alone.

Deep review finding S-1, roadmap DR-22.

## Changelog

- 🔒 **Security:** Redact Apprise `URLS`, Rocket.Chat `USER_ID`, and Telegram `CHATID` in debug dumps.
- 🔧 **Changed:** Resolve provider-specific sensitive fields by provider segment.
- ✨ **Added:** Tests for top-level fields, `env` pair values, and preserved `parse.urls` flags.


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed at bc7f9b6306681eb6dbc42ee0bff17cca318514e1. Data-only change to the provider-specific redaction map, tests cover the Apprise slack:// URL case.

ALARGECOMPANY (approved):
Reviewed at bc7f9b6306681eb6dbc42ee0bff17cca318514e1. Data-only change to the provider-specific redaction map, tests cover the Apprise slack:// URL case.

PR #960, opened 2026-09-03
feat(approvals): update approval queue with approve, reject and defer

scttbnsn:
Adds the update approval queue (CA-2). Updates that need a person now collect in a queue instead of only firing a notification, and a person can approve, reject, or defer each one from the API.

What's in it, in landing order:

- Classification. `classifyApprovalCandidate` sorts every update candidate into no-candidate, blocked, auto-dispatch, or queue, using the same policy the watcher already applies. Hard-blocked candidates never queue; soft-blocked ones queue with their hold reasons.
- Store and reconciler. A LokiJS collection for approval rows and a reconciler that runs on every watcher report: it supersedes a row when the candidate moves, resolves it when the container disappears or the update applies, and re-queues an expired deferral. Agent sync writes nothing.
- Read API. `GET /api/v1/approvals`, `/summary`, `/:id`, with `dd:approval-created` and `dd:approval-resolved` SSE frames.
- Decisions. `POST /api/v1/approvals/:id/approve|reject|defer`. Approve reserves the row with a compare-and-swap, then runs the exact code path `POST /containers/:id/update` runs (extracted into `container-update-dispatch.ts`, so parity is structural and a test pins it). Any blocker rolls the row back to its pre-reservation state. Reject is the existing Skip policy and defer is the existing snooze, both through one shared writer, so the queue and the container panel can't drift. A decision is refused with 409 if the row's candidate no longer matches the container's current candidate, if the row is already decided, or if the system is in notify mode (approve only). Three audit actions (`update-approved`, `update-rejected`, `update-deferred`) and a `dd:approval-decided` SSE frame.
- Request validation matches the OpenAPI schema: days is an integer 1 to 365, until is a full RFC 3339 instant in the future and at most 365 days out, note is a string, unknown keys are rejected.

Not in this PR: the bulk endpoint, the UI (route, sidebar badge, dashboard tile), and the docs pages. Those are the next slices.

Two things worth knowing. The queue's defer validates more strictly than the container panel's snooze, which still accepts date-only and past values; aligning them means changing the panel's behavior and is left for a follow-up. And a row that is both already decided and whose candidate has moved answers "candidate superseded" rather than "already decided", because the candidate check runs first so a refusal never writes.

Full app suite: 13772 tests, 100% coverage. An adversarial review found five issues in the decision slice (candidate binding, stale fields on re-decision, rollback snapshot, request validation, SSE string bounds); all five are fixed in the last commits with mutation checks proving the tests bite.

## Changelog

- ✨ Added approval model, LokiJS store, reconciliation, API routes, OpenAPI schemas, audit actions, and SSE lifecycle events.
- ✨ Added approve, reject, and defer decisions with candidate binding, compare-and-set reservation, live revalidation, rollback, and retention handling.
- 🔧 Shared manual update dispatch and update-policy writers between existing container actions and approval decisions.
- 🔧 Added controller-only reconciler startup and authenticated `/api/v1/approvals` routes.
- 🐛 Added validation for notes, defer schedules, timestamps, unknown properties, superseded candidates, and bounded SSE strings.
- 🐛 Added tests for store behavior, reconciliation, API decisions, parity, events, startup, and OpenAPI conformance.

## Concerns

- Verify the `app/index.ts` import path. The summary reports `./approotions/reconcile.js`, which appears to be a typo for `./approvals/reconcile.js`.
- Confirm that approval store initialization and retention timers do not run in agent-only or migration startup modes.
- Confirm that SSE payload truncation preserves valid UTF-16 or Unicode boundaries.
- Confirm that rollback restores the exact prior approval row, including decision, resolution, timestamps, and operation fields.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed the diff and the CI run, approving.

ALARGECOMPANY (approved):
Reviewed the diff and the CI run, approving.

PR #961, opened 2026-09-03
fix(updates): pin the pulled image digest for docker updates

scttbnsn:
Closes DR-32, the Docker-path half of the mutable-tag race that #952 fixed for Compose.

The Docker action trigger verified the signature of the configured tag before the pull, pulled the tag, then scanned and created the replacement from the tag again. A registry retag between any two of those steps meant the verified image, the scanned image, and the deployed image could all differ. Compose got a two-phase preflight in #952; this brings the ordinary Docker path to the same guarantee.

What changes:

- #952's five identity helpers move up from `Dockercompose.ts` into the `Docker` base class as protected methods, verbatim. Compose inherits them, no behavior change there.
- Right after the pull, the Docker executor captures the pulled repository digest and binds `repo:tag@sha256:...`. Signature verification (now deferred to post-pull, against the digest), the vulnerability scan, SBOM generation, the runtime-config compatibility check, and the container create all use that pinned value. The operation record's `targetImage` and the compose-file tag sync keep the operator's tag on purpose.
- Under a `required` binding policy a capture failure terminalizes the operation before any rename or create. Under `DD_SECURITY_AVAILABILITY_POLICY=warn` it writes a `security-scan-skipped` audit entry and skips the gate rather than scanning a mutable reference.
- Order on the Docker path is now pull, digest binding, verify, scan/SBOM, pre-update hook, prune, backup insert, config lookup, rename, create. The pre-update hook and the prune/backup step used to run before verification, so an unsigned candidate could prune cached images and insert a duplicate backup row before being rejected. They now run only after the gate passes.

One behavior change to know about: a pre-update hook abort now happens after the image is on disk and the scan has recorded its result, and it terminalizes the operation as failed instead of leaving the row to expire.

Tests: a deterministic retag regression (the registry repoints the tag after the first inspect and every later step still receives the original digest), strict call-order assertions, no-side-effect assertions for a signature block and a required-policy binding failure, and the warn-mode skip. Full app suite 13639 green, 100% on the four touched sources. Two adversarial review passes; the one remaining finding (an image ID carrying two digests for the same repository) is filed as a post-GA follow-up.

Backport to dev/v1.6 follows once this lands.

## Changelog

- 🔒 **Security:** Pin pulled Docker images to `repo:tag@sha256:...` before verification, scanning, SBOM generation, runtime checks, and container creation.
- 🔧 **Changed:** Move image identity helpers to the shared `Docker` base class.
- 🔧 **Changed:** Defer signature verification, pre-update hooks, pruning, and backups until after image binding and the security gate.
- 🔧 **Changed:** Support `required`, `optional`, and `disabled` image-binding policies.
- 🔧 **Changed:** Record `security-scan-skipped` audits under warning availability policy.
- 🐛 **Fixed:** Terminalize binding failures and hook aborts as failed operations before side effects.
- 🔧 **Changed:** Preserve existing Docker Compose lifecycle ordering.
- ✨ **Added:** Regression tests for digest binding, registry retags, policy handling, ordering, and side-effect prevention.
- ⚠️ **Breaking:** Update `postPullHook` callbacks and `ContainerUpdateExecutorDependencies` to support image identity and `PostPullHookOptions`.

## Concerns

- Update all external `postPullHook` implementations for the new callback signature.
- Confirm `bindPulledImageIdentity` uses existing Docker image parsing and repository-matching utilities where available.
- Confirm Docker and Docker Compose providers share the same lifecycle conventions after moving identity logic to `Docker`.
- Confirm warning-mode skipped scans are visible in all required audit and operation states.


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed the diff and the CI run, approving.

ALARGECOMPANY (approved):
Reviewed the diff and the CI run, approving.

PR #973, opened 2026-09-03
fix(portainer): bind the update gate and redeploy identity to the pulled digest

scttbnsn:
Roadmap DR-45. The Portainer action overrides the Docker update path, so it never reached the DR-32 digest binding: Update Bouncer and signature verification ran against the mutable tag, and the before/after image-id check read the tag twice, which an A -> B -> A retag passes.

What changes:

- Portainer calls the inherited `bindPulledImageIdentity` after the pull and passes the result to the post-pull hook, so the gate verifies and scans `repo:tag@sha256:...`. Policy follows `getPostPullIdentityBindingPolicy`: `required` fails closed before any Portainer PUT, warn mode records `security-scan-skipped` and still runs the pre-update hook and pre-runtime lifecycle.
- The redeploy identity comes from the bound digest's image id, not a re-read of the tag. Portainer performs the create, so this is not the Docker path's atomic binding; a retag between the PUT and Portainer's own create is detected by convergence against the scanned image id and the original stack file is restored. Pinning the digest into the operator's stack file was rejected: compose does not do it either by default, and `normalizeImageReference` would collapse the reference and break the next update's original-image guard.
- Stacks Portainer does not report as standalone Compose (`Type === 2`, checked on both the list entry and the detail response) are refused before fetch, pull, hook or PUT. Swarm workers pull the tag on their own nodes, so a local scan gates nothing there.
- `UpdateLifecycleExecutor` now fails an update whose runtime override deferred the gate but never invoked the post-pull hook. Four compose test stubs that resolved without calling the hook now call it, as the real refresh does.

Review: two Codex adversarial passes; the first found the tag re-read race and the missing stack-type check, both fixed here, plus a docs overstatement, reworded.

Verification: portainer suite 105 tests at 100% on Portainer.ts and UpdateLifecycleExecutor.ts; docker + dockercompose suites 1781 tests; biome and tsc clean; full pre-push gate green on push.

## Changelog

- 🔒 Security: Bind security gates to the digest pulled from the registry.
- 🔧 Changed: Enforce post-pull hook invocation for deferred lifecycle updates.
- 🔧 Changed: Compare redeploy convergence against the scanned digest image ID.
- 🐛 Fixed: Restore the original stack when redeployment converges on an unscanned image.
- ⚠️ Breaking: Reject Swarm, Kubernetes, and undeclared Portainer stack types before fetching or updating.
- ✨ Added: Add coverage for digest pinning, skipped scans, convergence failures, stack validation, and lifecycle-hook enforcement.
- 🔧 Changed: Update Portainer documentation for digest-bound verification and redeploy behavior.
- 🔧 Changed: Align Docker Compose test mocks with deferred post-pull lifecycle behavior.
- 🔧 Changed: Verification passes for Portainer, Docker, Docker Compose, Biome, TypeScript, and the pre-push gate.

## Concerns

- Update all `performContainerUpdate` callers to support the expanded `postPullHook` signature.
- Confirm that required security binding fails before any Portainer update operation.
- Confirm that non-standalone stack validation remains before stack-file retrieval and image pulling.


scttbnsn:
@coderabbitai review

scttbnsn:
Merge basis: all 11 required contexts green on ca6387349. CodeRabbit reviewed the code head 1cf2c61ad with no findings; the commits after it are test tightening (3bc1995eb), docs wording (266e08081) and the dev/v1.8 changelog merge (ca6387349), with no runtime change. Two adversarial review passes on the code head are recorded above; the residual convergence-window race they raised is tracked as DR-53 for its own PR.

biggest-littlest (approved):
Required checks green on ca6387349; code head reviewed, follow-ups are tests and docs only.

ALARGECOMPANY (approved):
Required checks green on ca6387349; code head reviewed, follow-ups are tests and docs only.

PR #976, opened 2026-09-03
test(ui): make the DashboardGrid resize spec deterministic

scttbnsn:
Roadmap DR-48. `DashboardGrid.spec.ts` "resizes with widget bounds and fallback bounds" flaked under full-suite load.

Root cause: Vue's DOM event invoker skips a listener when the event's `_vts` timestamp is not strictly after the listener's `attached` time, both millisecond-resolution `Date.now()`. The resizer button's `pointerdown` listener sits behind the grid item's capture-phase `pointerdown` listener on the same event, so a raw `dispatchEvent()` landing in the same millisecond as mount silently dropped `startResize()`. Fast runs hit it more often.

Fix: advance a fake clock past mount before the raw dispatches, so the event timestamp is unambiguously later regardless of run speed. No timeouts raised, no retries. Verified 30/30 sequential runs and one run concurrent with the rest of `tests/views`; biome clean. Test-only change, no changelog entry.

## Changelog

- 🐛 Fixed flaky `DashboardGrid` resize test timing.
- 🔧 Advanced fake timers before each raw `pointerdown`.
- 🔧 Restored real timers with `finally`.
- 🔧 Kept pointer sequences and layout assertions unchanged.

### DIFF
--- a/ui/tests/views/dashboard/DashboardGrid.spec.ts
+++ b/ui/tests/views/dashboard/DashboardGrid.spec.ts
@@ -241,19 +241,39 @@ describe('DashboardGrid', () => {
     } as DOMRect);
     const resizers = wrapper.findAll('.dd-grid-resizer');
 
-    resizers[0]!.element.dispatchEvent(pointerEvent('pointerdown', { clientX: 0, clientY: 0 }));
-    window.dispatchEvent(pointerEvent('pointermove', { clientX: 500, clientY: 200 }));
-    window.dispatchEvent(pointerEvent('pointercancel', {}));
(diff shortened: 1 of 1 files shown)


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
All required checks green on 949257604; CodeRabbit review finished with no inline findings.

ALARGECOMPANY (approved):
All required checks green on 949257604; CodeRabbit review finished with no inline findings.

PR #978, opened 2026-09-03
ci(secrets): scope the pull_request history scan to base..HEAD

scttbnsn:
The "Security: Secrets" job checks out with `fetch-depth: 0`, restores the scanner policy files from the PR's base ref, and then runs `scripts/scan-secrets.sh`, which called `gitleaks git --log-opts="--all"`. So on every pull request the git-history scan walked every branch ever pushed to the repo under the base branch's ignore file. Yesterday that meant two intermediate commits on one feature branch turned Secrets red on an unrelated v1.7 PR, and the branch could not fix its own finding because its ignore entries never applied. It was worked around by squashing the branch.

Change: on `pull_request` the workflow sets `SCAN_SECRETS_LOG_OPTS` to `origin/<base>..HEAD` and fetches that base ref into `refs/remotes/origin/<base>` first; on push, schedule and every other event it stays `--all`. The script reads the variable with `--all` as the default, so local runs are unchanged. The policy-restore step is untouched: policy still comes from the base ref. The script that runs on this PR is still the base's copy, so this PR itself gets the old full scan; the scoped scan takes effect for PRs opened after this merges.

Second commit: zizmor 1.30 (which the pre-push gate runs whenever a workflow file changes) flags the two `./.github/actions/wait-for-successful-branch-ci` references in release-cut.yml with its new self-repository audit. GitHub's `$/` form fixes zizmor but actionlint, via qlty, rejects it ("ref is missing"), so the references keep the workspace-relative form with an inline `# zizmor: ignore[self-repository]` until actionlint accepts `$/`. Without that, no workflow change can get through the gate.

Tests: `.github/tests/ci-verify-workflow.test.ts` pins the per-event log-opts and the script's env var handling.

Roadmap item DR-51.

## Changelog

- ✨ Added pull-request-scoped Secrets history scans using `origin/<base>..HEAD`.
- 🔧 Changed `scan-secrets.sh` to read `SCAN_SECRETS_LOG_OPTS`, defaulting to `--all`.
- 🔧 Changed CI to fetch the pull-request base ref before scanning.
- 🔧 Added zizmor ignores for two workspace-relative actions in `release-cut.yml`.
- 🔒 Overrode `fflate` to `0.8.3` for CVE-2026-45820.
- ✨ Added workflow tests for event-specific scan options and environment-variable handling.

## Concerns

- Verify that the relevant lockfile includes the `fflate` override.
- Confirm that pull-request workflows have permission to fetch the base ref with complete history.
- Confirm that the zizmor annotations suppress only the intended findings.


scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Read the base..HEAD scoping against the workflow and its test; matches the description.

ALARGECOMPANY (approved):
Read the base..HEAD scoping against the workflow and its test; matches the description.

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

scttbnsn:
@coderabbitai review

PR #989, opened 2026-09-03
chore(release): prepare v1.7.0-rc.9

scttbnsn:
Release prep for v1.7.0-rc.9 on dev/v1.7, mirroring the rc.8 prep (#963).

What moves under the rc.9 heading in CHANGELOG.md: the fflate CVE override, the single-flight `watchFromCron()` fix with its scan deadline, the `once=true` history-key fix, the deprecation banner corrections, the marketing-site Get Started fix and the docs audit corrections (#979, #988). The rc.9 highlights block is added to README.md and the six translations, the RC identity string is bumped in the demo mocks, site config, API docs and quickstart, and the release identity tests, changelog-links test and pre-cut check test track rc.9.

`node scripts/release-precut-check.mjs v1.7.0-rc.9` passes. Workflow tests, scripts tests and apps/web script tests pass.

After this merges: reconcile dev/v1.7 with main, open the promotion PR, cut `v1.7.0-rc.9` from main.

PR #990, opened 2026-09-03
chore(release): promote v1.7.0-rc.9 to main

scttbnsn:
Promotes the exact final dev/v1.7 tree after #989 so v1.7.0-rc.9 can be cut from protected main.

The history-only ours merge (`ebc21e208`) connects the rc.8 squash promotion without changing the dev/v1.7 tree. main has no commits since that promotion. Verified byte-for-byte tree equality of the reconcile commit against origin/dev/v1.7, main ancestry, and the full pre-push gate.

After merge: verify origin/main and origin/dev/v1.7 have identical trees, confirm the push CI on main, then dispatch release-cut.yml for v1.7.0-rc.9.


biggest-littlest (approved):
Tree-equal promotion of dev/v1.7 for the rc.9 cut.

ALARGECOMPANY (approved):
Tree-equal promotion of dev/v1.7 for the rc.9 cut.

PR #999, opened 2026-09-04
fix: reserve the once slot on batch and digest paths, clear the cron deadline on deregistration (1.6)

scttbnsn:
Backport of #997 to the 1.6 line, cherry-picked from 680eee12d. All six code files applied clean; only CHANGELOG conflicted because 1.6's Unreleased Fixed section has its own bullets, resolved by appending the new one.

What it fixes, same as on 1.8:

- Batch and digest `once` eligibility reserve the slot instead of reading history, so a manual scan overlapping a cron scan can no longer send the same candidate twice in batch or digest mode. Reservations are tracked in their own collection and released in a finally; the digest flush reserves exactly what it sends and drops anything already notified; retry-buffer containers reserve too. Both new sites are gated on `once` like the existing ones, so `once=false` behaviour is unchanged.
- The cron-scan deadline timer is cleared on deregistration, the race settles with a deregistration sentinel without logging, a late cleanup cannot clear a newer scan's handle, and a scan requested after deregistration (the pending docker-event debounce, the startup and discovery timers) returns without starting.

16 new tests, all failing against the previous code. Reviewed on #997 by Codex (three findings fixed there) and CodeRabbit (one real finding fixed, one refuted and filed as DR-72).

Three comments that named the discovery-settle timer were reworded because 1.6 has no pending-discovery path (dropped from #991 for the same reason); no code hunk was dropped. 11 of the 16 new tests fail against the 1.6 base, the other five are the deliberate still-works guards.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Cherry-pick of #997 (reviewed there by Codex and CodeRabbit); code hunks identical to 1.8, only three comments reworded for the missing discovery-settle path. CodeRabbit is rate limited on this one.

ALARGECOMPANY (approved):
Cherry-pick of #997 (reviewed there by Codex and CodeRabbit); code hunks identical to 1.8, only three comments reworded for the missing discovery-settle path. CodeRabbit is rate limited on this one.

PR #1007, opened 2026-09-04
refactor(triggers): fold the once gate into reserveOnceNotificationSlot

scttbnsn:
Roadmap DR-73, follow-up to #997. `reserveOnceNotificationSlot` opened with a plain history read and `recordNotifiedForResult` writes history regardless of `once`, so every caller had to carry its own `if (!this.configuration.once)` guard or it would suppress repeat sends for `once=false` triggers. #997 added two more such sites and CodeRabbit caught one missing guard there. This moves the check into the helper: with `once` off it returns true without reserving, the five call-site guards are gone, and the release in each finally stays a no-op for never-reserved keys.

Behaviour is unchanged for both settings. Two tests drive `handleContainerReport` concurrently and assert that `once=false` never adds a key to `inFlightOnceNotificationKeys` while `once=true` does.

## Changelog

- ✨ Added concurrent tests for `once=false` and `once=true`.
- 🔧 Moved the `once` check into `reserveOnceNotificationSlot`.
- 🔧 Removed redundant guards at reservation call sites.
- 🔧 Kept release calls safe for unreserved keys.
- 🐛 Prevented duplicate concurrent notifications when `once=true`.

## Concerns

- Verify reservation cleanup when notification sending fails.
- Verify all call sites use the centralized reservation helper.
- Check providers for inconsistent reservation behavior.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Approving: pure refactor, adversarial review found no correctness defect and the nits are folded into the second commit, checks green.

ALARGECOMPANY (approved):
Approving: pure refactor, adversarial review found no correctness defect and the nits are folded into the second commit, checks green.

PR #1009, opened 2026-09-04
fix(rollback): prefer the backup digest when pinning a rollback image

scttbnsn:
Two roadmap items from the #984 follow-up list.

DR-64, fixed. `selectPulledRepoDigest` used `container.result.digest` as the tie-break when the daemon records several `RepoDigests` for the pulled repository. On the update path that is the manifest the watcher resolved for the candidate, which is right. On a manual rollback the pulled image is the backup's retained image, and the candidate's digest describes nothing about it, so with two RepoDigests for the same repository the rollback could pin to the manifest it was undoing. `bindPulledImageIdentity` now takes a `preferredDigest` option, `resolveRollbackImageReference` passes the backup record's captured digest (and nothing else), and the update path is unchanged. The end-to-end test with two RepoDigests fails without the fix.

DR-66, refuted. The ticket said compose-file-once replicas skip the deferred pre-update hook and prune/backup because no post-pull hook is passed. That hook is not the carrier: `createTriggerContext` sets `deferPreRuntimeUpdateLifecycle` false when the gate already completed, so the lifecycle executor runs hook, prune and backup eagerly before `performContainerUpdate`. A live probe produced preHook, prune, backup, refresh, postHook, and an existing test already asserted it. The second commit pins the per-replica rate (two replicas, one gate each in the preflight, hook and prune/backup once each) so the claim stays refuted. The smaller ordering defect found on the way (eager lifecycle fires before the refresh validates the container) is tracked separately as DR-75.

One thing to know when reading `app/util/backup.ts`: `{ preferredDigest: backup.imageDigest }` is always undefined on that path today because a record with a digest returns early above it. It is written that way so the option stays correct if that early return is ever narrowed; the comment there says so.

## Changelog

- ✨ Added rollback digest tie-breaking with `preferredDigest`.
- 🐛 Fixed rollback image selection when multiple `RepoDigests` exist.
- ✨ Added rollback tests for ambiguous digests and digest-resolution policies.
- ✨ Added compose-file-once tests for per-replica hooks, pruning, backups, and refreshes.
- 🔧 Preserved update-path digest selection behavior.

## Concerns

- Verify that rollback records with a captured digest pass that digest to `bindPulledImageIdentity` before any early return.
- Address the lifecycle ordering issue tracked by DR-75.
- Check all `RollbackImageIdentityBinder` implementations and call sites after the interface change.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed against the Opus adversarial pass: the pinned-digest tie-break now fails closed under the required binding policy, and the DR-66 test pins the eager lifecycle path.

ALARGECOMPANY (approved):
Reviewed against the Opus adversarial pass: the pinned-digest tie-break now fails closed under the required binding policy, and the DR-66 test pins the eager lifecycle path.

PR #1015, opened 2026-09-04
fix(agent): guard the 1.6 watch path prune on an empty list

scttbnsn:
Backport of the watch-path half of #922 to the 1.6 line.

`AgentClient.watch()` on `dev/v1.6` still ingested the agent's report list and then pruned, with no empty-list guard. A manual recheck during an agent hiccup handed the controller an empty list, so every container the watcher owned was deleted from the store along with its stored update policy (the #565 symptom). The rc.7 CHANGELOG bullet also claimed `watch()` was already guarded, which was not true on this line.

`watch()` now builds the container list first, prunes before ingest so the same-identity policy stash exists before `insertContainer()` consumes it, and skips the prune entirely on an empty list, logging a warning when the agent had connected before. Three tests pin the guard and the ordering; each fails against the pre-fix code. The rc.7 bullet is corrected and a v1.6.1 Fixed bullet describes the change.

The other three ingest paths on 1.6 (handshake, watcher snapshot, container sync) carry the empty-list guard from rc.7 but still prune after ingest. That reorder is tracked separately; it touches the snapshot event ordering and belongs in its own reviewed change.

Fixes: #565


biggest-littlest (approved):
Reviewed the watch-path port against #922 and the Opus verification (three pinning tests, ordering trace, no regression on non-empty lists).

ALARGECOMPANY (approved):
Reviewed the watch-path port against #922 and the Opus verification (three pinning tests, ordering trace, no regression on non-empty lists).

Issue #939, opened 2026-08-29
SSO - Authentik not working with v1.7.0-rc.6


begunfx:
### Drydock version

v1.7.0-rc.6

### What happened?

Tried to login to drydock via sso/authentik.  It just returns me to the login screen.

### What did you expect?

See above

### Steps to reproduce

See above

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

24.0.1

### OS / Architecture

Synology DSM 7

scttbnsn:
Thanks for the report. There is not enough configuration or callback evidence yet to distinguish a lost session cookie from an Authentik client mismatch. Please share redacted DD_PUBLIC_URL, DD_SERVER_COOKIE_SAMESITE, DD_SERVER_TRUSTPROXY and TLS settings, the configured callback URI, the callback response status/body, whether the drydock.sid cookie is sent on the callback, and Drydock logs around that request. For a cross-origin Authentik flow, SameSite=lax is required.

begunfx:
> Thanks for the report. There is not enough configuration or callback evidence yet to distinguish a lost session cookie from an Authentik client mismatch. Please share redacted DD_PUBLIC_URL, DD_SERVER_COOKIE_SAMESITE, DD_SERVER_TRUSTPROXY and TLS settings, the configured callback URI, the callback response status/body, whether the drydock.sid cookie is sent on the callback, and Drydock logs around that request. For a cross-origin Authentik flow, SameSite=lax is required.

Where should I look for the callback response status/body?  I assume the other asks are in the docker stack.

begunfx:
I went all the way back to 1.6.0 and still had the issue.  After some help from AI it turned out to be a cookie id issue.  I cleared the cache specfically for my drydock sub-domain and was then able to sign-in again via Authentik.  I think you can close this.  Thanks!

scttbnsn:
Thanks for confirming. Since clearing the stale browser state fixed the Authentik flow and you asked us to close this, I'm closing it as resolved.

depuits:
I seem to have the same issue using 1.7.0-rc10. @begunfx did clearing the cache fix this issue on 1.7.0 or are you now running 1.6? For me clearing the site cache/data fixed it when going back to 1.6.1-rc7 but not on while on version 1.7.0-rc10.

scttbnsn:
[...]

PR #1016, opened 2026-09-04
fix(ui): keep the service worker off server-owned routes so the OIDC callback reaches the app

scttbnsn:
Since v1.7.0-rc.1 the UI ships a vite-plugin-pwa service worker. Its navigation fallback only denied `/api/`, so every other top-level navigation was answered from the precached app shell. The identity provider redirects the browser to `/auth/oidc/<name>/cb?code=...`, which is a document navigation, so the callback got `index.html` and the code exchange never reached Express. Login bounced straight back to the login page (#939, reproduced by two users on rc.10).

The denylist now covers every path Express owns on the SPA origin: `/api`, `/auth/`, `/health` and `/metrics`, exported as `SERVER_OWNED_NAVIGATION_PATTERNS` so the patterns can be asserted against real callback URLs in tests. Bare `/auth` stays with the SPA because it is a Vue route. No `runtimeCaching` rule was added for `/auth`: workbox only calls `respondWith` when a route matches, so the browser performs the callback request itself and its 302 is never touched by the worker.

Users stuck on rc.10 pick up the new worker on their next load: `sw.js` is served with `Cache-Control: no-cache` and `registerType: 'autoUpdate'` sets `skipWaiting` and `clientsClaim`.

Fixes: #939

Also carries the cherry-pick of #976 (`e10569df9`), the deterministic DashboardGrid resize spec from dev/v1.8. The old spec fails whenever the resizer's pointerdown lands in the same millisecond Vue attached its listener, which happens about half the time on a warm test run, and it tripped this branch's pre-push gate once already. Test-only, no product change.


biggest-littlest (approved):
Reviewed against the Opus verification of the route table, workbox matching semantics and the self-heal path; denylist now pinned to the Express mounts.

ALARGECOMPANY (approved):
Reviewed against the Opus verification of the route table, workbox matching semantics and the self-heal path; denylist now pinned to the Express mounts.

PR #1025, opened 2026-09-04
fix(registry): prune controller-owned records when no local watcher is configured

scttbnsn:
Fixes the stranded-container case from the DR-104 review: with `DD_LOCAL_WATCHER=false` and no `DD_WATCHER_*` configured, `pruneOrphanedLocalContainers()` early-returned because there were no local watcher names, so a record the controller used to own (no `agent` field) survived every restart. When an agent then reported the same container, the ownership gate refused it as controller-owned, and the container was stuck on the old record with no owner that would ever refresh it.

The prune now runs when no local watcher registration was attempted at all: every controller-owned record is orphaned by definition, so it is deleted and the agent's next report is accepted. The early return is kept for the transient case where local watchers were configured but every registration failed (unreadable CA file, bad socket, config rejected), because wiping records over a bad startup is the wrong trade. The count comes from the actual registration promise array, so it cannot drift from what was attempted. The comment in the code is explicit that this only covers the all-failed case; a partial failure (two watchers, one down) still prunes the failed watcher's records exactly as before, which is a separate change.

Operator-visible side effect, stated in the CHANGELOG and on both docs pages that describe `DD_LOCAL_WATCHER`: the pruned record's UI-set update policy overrides (snooze, maturity mode, skipped tags and digests) go with it. The prune deletes without a replacement stash, and the retention key is agent plus watcher plus name, so a move to an agent could not inherit them anyway.

Tests: a controller-owned record is pruned and an agent-owned one on the same watcher name survives when nothing local is configured; a record owned by a configured watcher that failed to register is kept even with the flag off. The existing "skip when no local watchers registered" test still passes unchanged. 446 files, 15039 tests, coverage 100% on all four metrics.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed with an adversarial Opus pass: registerWatchers now reports attempted local registrations, prune skips only when a local watcher was configured and none registered, CHANGELOG discloses the lost UI overrides, docs sentence on both pages. Coverage 100%, gate green.

ALARGECOMPANY (approved):
Reviewed with an adversarial Opus pass: registerWatchers now reports attempted local registrations, prune skips only when a local watcher was configured and none registered, CHANGELOG discloses the lost UI overrides, docs sentence on both pages. Coverage 100%, gate green.

PR #1026, opened 2026-09-04
fix(compose): run the once-path hook, prune and backup after the refresh checks

scttbnsn:
DR-75 from the roadmap, found while refuting DR-66.

On the compose-file-once path (post-pull gate already completed in the preflight), the eager lifecycle ran the operator pre-update hook, the image prune and the rollback backup row before `refreshComposeServiceWithDockerApi` had checked the container. The ordinary compose path runs all three inside the refresh, after `getCurrentContainer` and `ensureComposeRuntimeState`. So a once-path refresh that aborted because the container had vanished, or because its inspect came back without usable runtime state, had already fired the hook, pruned images and written a backup for an update that never happened.

The change makes the once-path use the same order as the ordinary path. `createTriggerContext` always defers the pre-runtime lifecycle, `performContainerUpdate` always forwards the post-pull hook, and when the preflight has already gated the image the hook is forwarded with `skipSecurityGate` so the security gate still runs exactly once. The eager path existed because no hook was passed down at all on the gated path, which would have tripped the executor's missed-hook guard; forwarding a gate-suppressed hook is what makes deferral possible. Self-update and infrastructure updates keep the eager order (`!isSpecialUpdate`) and never reach `performContainerUpdate`.

Tests: two new once-path abort tests (vanished container, inspect without runtime state) assert the reject message, that the preflight gate ran once, and that no hook, no prune and no backup row happened. The ordering test in the extended suite now pins hook, prune and backup after the check. The stand-in compose mocks call a shared helper so they forward the hook's real arguments the same way the test helper does. 446 files, 15040 tests, coverage 100% on all four metrics.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Reviewed with an adversarial Opus pass: pre-runtime lifecycle always deferred through the trigger context, post-pull hook forwarded once and the security gate skipped only after it ran, abort paths for vanished and non-running containers covered. Coverage 100%, gate green.

ALARGECOMPANY (approved):
Reviewed with an adversarial Opus pass: pre-runtime lifecycle always deferred through the trigger context, post-pull hook forwarded once and the security gate skipped only after it ran, abort paths for vanished and non-running containers covered. Coverage 100%, gate green.

biggest-littlest (approved):
Reviewed with an adversarial Opus pass: pre-runtime lifecycle always deferred through the trigger context, post-pull hook forwarded once and the security gate skipped only after it ran, abort paths for vanished and non-running containers covered. Rebased onto #1025; gate green, Cucumber flake rerun clean.

ALARGECOMPANY (approved):
Reviewed with an adversarial Opus pass: pre-runtime lifecycle always deferred through the trigger context, post-pull hook forwarded once and the security gate skipped only after it ran, abort paths for vanished and non-running containers covered. Rebased onto #1025; gate green, Cucumber flake rerun clean.

PR #1028, opened 2026-09-05
fix(rollback): pull the backup image only when it is not already local

scttbnsn:
Follow-up to #1020, found in the adversarial review of the 1.6 backport (#1027).

`Docker.pullRollbackImage` runs before any destructive step in `HealthMonitor.performRollback`, and `pullImage` throws on any registry failure. On an airgapped host, or under a Docker Hub anonymous 429, that turned a rollback that used to succeed from the retained local image into "Auto-rollback failed" with the unhealthy container left running. The pull is redundant in exactly that case, and the compose refresh's `requireLocalImage` refusal already covers the genuinely missing image.

`pullRollbackImage` now inspects the rollback reference locally first (`getImage(ref).inspect()`; any failure counts as absent) and skips the pull with an info log when it is on the host. Registry resolution and `getAuthPull()` only run on the pull path, so an airgapped host never touches the network. A pull failure still propagates, now as "Backup image <ref> is neither on the host nor pullable". `performRollback`'s log line says "resolving backup image" instead of "pulling".

On 1.8 the reference is the hybrid `repo:tag@sha256:...` from `buildRollbackImageReference`; that form is already inspected by `verifyPulledImageCompatibility` under `requireLocalImage` and by `capturePulledImageIdentity`, so the skip decision and the later verification look at the same reference.

Tests: image local (no pull, no auth lookup, info log), inspect 404 then pull with the resolved auth, inspect non-404 error then pull, inspect 404 and pull rejects with the new message, plus an end-to-end HealthMonitor test with a real Dockercompose trigger where `getAuthPull` is set to reject and the rollback still succeeds from the local image. 446 files, 15041 tests, coverage 100% on all four metrics. Same change is on 1.6 inside #1027; the 1.7 port follows.


scttbnsn:
@coderabbitai review

biggest-littlest (approved):
Port of the 1.6 DR-114 change reviewed with an adversarial Opus pass: local inspect before the pull, registry resolution and getAuthPull behind the same check, pull failure rethrown with cause, end-to-end HealthMonitor test with a rejecting getAuthPull. Gate green, 100% coverage.

ALARGECOMPANY (approved):
Port of the 1.6 DR-114 change reviewed with an adversarial Opus pass: local inspect before the pull, registry resolution and getAuthPull behind the same check, pull failure rethrown with cause, end-to-end HealthMonitor test with a rejecting getAuthPull. Gate green, 100% coverage.

PR #1032, opened 2026-09-05
ci(release): verify base image indexes and platform binaries at the cut (1.6)

scttbnsn:
Script-only backport of the DR-111 release guards from #1024 (dev/v1.7). release-cut.yml checks out `source_ref` for a maintenance cut and calls `scripts/check-dockerfile-base-indexes.sh` and `scripts/check-image-arch.sh` from that tree, so dev/v1.6 needs the scripts before its next cut or the cut fails on a missing file.

No Dockerfile change: 1.6's `node:24-alpine` and `alpine:3.24` pins are already image indexes (verified against the registry by the guard itself), and the arm64 bug in drydock#1021 never shipped on this line. The two ported workflow tests were adapted to 1.6's workflow shape (plain `docker/build-push-action`, no retry-wrapped buildx step) in a separate commit so the three cherry-picks stay identical to their 1.7 source.

Verified: guard passes on the 1.6 Dockerfile, 179 script tests, 97 workflow tests, biome and qlty clean.

### DIFF
--- a/.github/tests/ci-verify-workflow.test.ts
+++ b/.github/tests/ci-verify-workflow.test.ts
@@ -518,5 +518,7 @@ test('build job checks base image pins before building, for the platforms it smo
   // digest resolves the same for every --platform, so the arm64 stage builds
   // green on an amd64 rootfs (#1021). Both must cover the same platforms.
   expect(smokePlatforms).toBe('linux/amd64,linux/arm64');
-  expect(guard?.run).toBe(`scripts/check-dockerfile-base-indexes.sh Dockerfile ${smokePlatforms}`);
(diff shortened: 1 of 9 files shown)


biggest-littlest (approved):
Scripts-only backport of the #1024 guards so a 1.6 maintenance cut finds them in the checked-out tree; no Dockerfile change, 1.6 pins verified as indexes. Gate and CI green.

ALARGECOMPANY (approved):
Scripts-only backport of the #1024 guards so a 1.6 maintenance cut finds them in the checked-out tree; no Dockerfile change, 1.6 pins verified as indexes. Gate and CI green.

PR #1033, opened 2026-09-05
fix(registry): keep policy overrides and configured watchers' records on prune

scttbnsn:
DR-112 and DR-113, both found while fixing DR-108 (#1025).

DR-112: a container handed from the controller's local watcher to an agent came up with its update policy cleared. The startup prune deleted the controller-owned record with a plain `deleteContainer`, so the policy stash never ran, and the stash key is identity-based (`agent::watcher::name`) so a controller-to-agent move could not consume it anyway. The prune now deletes with `identityChangeExpected`, which stashes the policy under a second key space keyed on the Docker container id (`id::<containerId>`), the one field that survives a move or a watcher rename. `restoreRetainedUpdatePolicy` tries the identity key first and falls back to the id key. Name-based matching was rejected: two hosts can both run `myapp`. A container recreated as part of the migration mints a new id and matches neither key; the CHANGELOG says so.

DR-113: with two configured local watchers where one failed to register, the prune kept only the survivor's names and deleted the failed watcher's records. The keep-set is now registered plus configured names. Fixing that exposed a race: `registerWatchers` used `Promise.all`, which settles on the first rejection while the other registrations are still in flight, so a healthy watcher could land in the registry after the prune had read it. It now uses `allSettled` and rethrows the first rejection, keeping the existing warn path.

The #1025 CHANGELOG bullet and both docs pages no longer claim the overrides are lost. Left out on purpose, same shape, own items: `pruneOrphanedAgentContainers` (agent removed from config) and `AgentClient.pruneOldContainers` (container moving between two agents) still delete without the stash.

Tests: 446 files / 15053 passing, 100% on every touched file, biome and tsc clean.


biggest-littlest (approved):
Policy stash keyed on the Docker id for controller-to-agent moves and watcher renames, configured watcher names kept on a partial registration failure, allSettled closes the prune race. 446 files / 15053 tests, 100% on every touched file, gate and CI green.

ALARGECOMPANY (approved):
Policy stash keyed on the Docker id for controller-to-agent moves and watcher renames, configured watcher names kept on a partial registration failure, allSettled closes the prune race. 446 files / 15053 tests, 100% on every touched file, gate and CI green.

PR #1035, opened 2026-09-05
fix(agent): gate an agent-owned container's auto-install by its maintenance window

scttbnsn:
DR-96. Agent-owned containers were never maintenance-window-gated on the controller, while the container panel said their install was deferred. `AgentWatcher` implemented no `isMaintenanceWindowOpen`, so `resolveWatcherMaintenanceWindowOpen` failed open and `Trigger.deferAutoUpdateForMaintenanceWindow` bailed; the UI field fell back to the agent's masked `maintenancewindowopen` and reported a deferral nothing enforced.

`AgentWatcher` now answers `isMaintenanceWindowOpen()` the same way a Docker watcher does: under the controller Docker transport it proxies to the live delegate watcher; otherwise it reads the agent's latest watcher snapshot (`AgentClient.getWatcherSnapshot(...).configuration.maintenancewindowopen`, refreshed on every agent scan and already feeding `/api/watchers`), falls back to the handshake configuration, and fails open when an older agent reports nothing, which keeps pre-fix agents exactly as ungated as before and stops the UI claiming a deferral. No poll-on-window-open was added: under the default `install` scope the agent's cron keeps scanning and reporting, so the next report re-evaluates the deferred container through the normal pipeline.

The manual agent scan (`app/agent/api/watcher.ts`) is left ungated on purpose; local manual scans are ungated too, the scan-scope gate only ever lived in the cron wrapper. Docs: four pages no longer say agent watchers are not deferred yet; CHANGELOG amended plus a Fixed bullet.

Tests: six `isMaintenanceWindowOpen` cases on AgentWatcher (delegate, snapshot, handshake fallback, absent field, unknown agent, no agent). AgentWatcher.ts 100%. 446 files / 15062 tests, biome and build clean.


biggest-littlest (approved):
AgentWatcher.isMaintenanceWindowOpen proxies the delegate under controller transport and otherwise reads the agent snapshot, so agent-owned containers defer exactly like local ones and the UI no longer claims a deferral that is not enforced. Coverage 100%, gate and CI green.

ALARGECOMPANY (approved):
AgentWatcher.isMaintenanceWindowOpen proxies the delegate under controller transport and otherwise reads the agent snapshot, so agent-owned containers defer exactly like local ones and the UI no longer claims a deferral that is not enforced. Coverage 100%, gate and CI green.

PR #1036, opened 2026-09-05
fix(agent): keep update policy across agent moves and reconnect recreates (1.6)

scttbnsn:
Two 1.6 fixes in the agent ingest path.

DR-112/113 port (cherry-pick of #1033 from dev/v1.8): a container handed from the controller's local watcher to an agent keeps its update policy, via a second policy stash keyed on the Docker container id (`id::<id>`) written when a record is deleted for an identity change; one local watcher failing to register no longer deletes its records when another registers (keep-set is registered plus configured names, and `registerWatchers` uses `allSettled` so a late success cannot land after the prune). This also closes the `Promise.all` race CodeRabbit flagged on #1030. The #1030 CHANGELOG bullet and both docs pages no longer claim the overrides are lost.

DR-103: three 1.6 ingest paths still pruned after ingest, so a container recreated while the agent was disconnected lost its update policy. `insertContainer()` restores the same-identity stash as its first statement, and only `deleteContainer(id, { replacementExpected: true })` writes it, but `_doHandshake()`, `handleWatcherSnapshotEvent()` and `handleContainerSync()` in `AgentClient` all ingested first and pruned after, so the stash was written after its only consumer ran and the replacement landed on the global default policy (the #565 symptom on reconnect). #922 reordered all three on 1.7; DR-99 ported only `watch()`. All three now prune before ingest, with a call-order and policy-retention test for each, a zero-container guard test on the snapshot and watch paths, and an audit-subscription test that the lifecycle row describes the surviving container.

DR-105 (deterministic DashboardGrid resize spec) needed no commit: 1.6's spec is already byte-identical to the #976 version and passes three runs in a row.

AgentClient.ts 100% on all four metrics; 401 files / 12871 tests, coverage 100% globally, biome and build clean.


biggest-littlest (approved):
1.6 backports of the id-keyed policy stash (DR-112), the registered-plus-configured keep-set with allSettled registration (DR-113) and prune-before-ingest on handshake, snapshot and sync (DR-103). Coverage 100%, gate and CI green.

ALARGECOMPANY (approved):
1.6 backports of the id-keyed policy stash (DR-112), the registered-plus-configured keep-set with allSettled registration (DR-113) and prune-before-ingest on handshake, snapshot and sync (DR-103). Coverage 100%, gate and CI green.

PR #1037, opened 2026-09-05
fix(registry): port the agent policy and prune fixes to 1.7

scttbnsn:
Ports to 1.7 plus the REL-2 docs carry-over, one commit each.

DR-108 (cherry-pick of #1030, itself #1025): with `DD_LOCAL_WATCHER=false` the startup prune now clears controller-owned records, so a container that moved to an agent stops being refused; the prune is still skipped when local watchers were configured and all failed to register.

DR-112/113 (cherry-pick of #1033): a container handed from the controller to an agent keeps its update policy through a second stash keyed on the Docker container id (`id::<id>`), written by `deleteContainer(id, { identityChangeExpected: true })`; one local watcher failing to register no longer deletes its records when another registers (keep-set is registered plus configured names, `registerWatchers` on `allSettled` so a late success cannot land after the prune). The DR-108 bullet and both docs pages no longer claim the overrides are lost.

REL-2 carry-over: the agents page's registries example mounts the CA file it points `DD_AGENT_REMOTE1_CAFILE` at and drops the plain-HTTP quickstart comment, hand-ported from #1014.

DR-105 needed no commit: 1.7's DashboardGrid spec is already the deterministic #976 version (it came in with #1016).

`app/registry/index.ts` and `app/store/container.ts` 100% on all four metrics; 421 files / 13748 tests, biome clean in app and ui, build clean.


biggest-littlest (approved):
1.7 ports of the controller-owned prune fix (DR-108), the id-keyed policy stash (DR-112), the registered-plus-configured keep-set with allSettled registration (DR-113), and the REL-2 docs carry-over. Coverage 100%, gate and CI green.

ALARGECOMPANY (approved):
1.7 ports of the controller-owned prune fix (DR-108), the id-keyed policy stash (DR-112), the registered-plus-configured keep-set with allSettled registration (DR-113), and the REL-2 docs carry-over. Coverage 100%, gate and CI green.

PR #1043, opened 2026-09-05
chore(security): allow the shifted webhook doc curl examples ahead of #1038

scttbnsn:
Additive only. #1038 adds two lines to content/docs/current/configuration/webhooks/index.mdx above the curl examples, which moves the four path-only curl-auth-header fingerprints from 46/115/122/129 to 48/117/124/131. The Secrets job checks out .gitleaksignore from the base branch, so a branch-side edit can't fix its own scan: the new line numbers have to be on dev/v1.8 first. This adds them next to the existing entries; the old ones stay until #1038 lands.


biggest-littlest (approved):
Additive .gitleaksignore entries for the line shift #1038 introduces; the Secrets job reads this file from the base branch. Gate and CI green.

ALARGECOMPANY (approved):
Additive .gitleaksignore entries for the line shift #1038 introduces; the Secrets job reads this file from the base branch. Gate and CI green.

PR #1044, opened 2026-09-05
test(rollback): pass backupScope to every health monitor case

scttbnsn:
DR-117, test only. `HealthMonitorOptions.backupScope` is required, but 20 of the 23 `startHealthMonitor` calls in HealthMonitor.test.ts omitted it. Tests are excluded from tsc and none of those cases reach the rollback path, so nothing changed at runtime; the calls now match the option type. Raised by CodeRabbit on #1041. The 1.8 port is a separate edit (19 calls there, one already fixed).

### Changelog

- 🔧 Changed all 20 relevant `startHealthMonitor` test calls in `HealthMonitor.test.ts` to include `backupScope`.
- 🔧 Used `{ containerName: 'test-container', includeLegacy: true }` for each call.
- 🐛 Fixed test call sites that omitted the required `HealthMonitorOptions.backupScope`.
- 🔒 Runtime behavior is unchanged.

### Concerns

- Confirm that the remaining 3 `startHealthMonitor` calls intentionally do not require `backupScope`.
- Verify the 1.8 port separately.


biggest-littlest (approved):
Test-only: every startHealthMonitor case passes backupScope. Gate and CI green.

ALARGECOMPANY (approved):
Test-only: every startHealthMonitor case passes backupScope. Gate and CI green.

PR #1047, opened 2026-09-05
fix(release): probe each platform by its own manifest digest in the image arch check (1.6)

scttbnsn:
1.6 port of the DR-118 fix (1.7 branch commit 1b439dfb3). The arch probe added in #1032 runs `docker run --platform <p> <repo>@<index digest>` once per platform; the classic image store keeps one variant per digest, so the arm64 pass dies with `docker: cannot overwrite digest` after amd64 passes. That is what failed the v1.6.1-rc.9 cut (run 33945788381). The script now resolves the platform's own manifest digest from `docker buildx imagetools inspect --raw` and runs that. The rc.9 cut is re-dispatched once this lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
Port of #1046. Per-platform manifest digest resolved from the index before the probe, script tests 24/24, gate green, only Vercel quota and qlty minutes noise on checks.

ALARGECOMPANY (approved):
Port of #1046. Per-platform manifest digest resolved from the index before the probe, script tests 24/24, gate green, only Vercel quota and qlty minutes noise on checks.

PR #1049, opened 2026-09-05
fix(release): probe each platform by its own manifest digest in the image arch check

scttbnsn:
Forward-port of #1046 to dev/v1.8.

The arch check ran `docker run --platform <p> <repo>@<index digest>` for each platform. On the classic image store the second platform fails with `docker: cannot overwrite digest` because the index digest is already bound to the first platform's image. This resolves each platform's own manifest digest from the index via `docker buildx imagetools inspect --raw` and runs that instead. Non-index references run unchanged. Attestation entries (os unknown) are skipped, zero or multiple matches error out before any docker run.

Script tests 20/20, workflow tests 260/260, CHANGELOG bullet under Unreleased / Fixed.


biggest-littlest (approved):
Port of #1046. Per-platform manifest digest resolved from the index before the probe, script tests 20/20, workflow tests 260/260, gate green, only Vercel quota and qlty minutes noise on checks.

ALARGECOMPANY (approved):
Port of #1046. Per-platform manifest digest resolved from the index before the probe, script tests 20/20, workflow tests 260/260, gate green, only Vercel quota and qlty minutes noise on checks.

PR #1051, opened 2026-09-05
fix(rollback): resolve auto-rollback through the identity binding policy

scttbnsn:
Three roadmap items plus the CI-11 header fix.

- DR-59: the health-check auto-rollback resolves its backup image through `resolveRollbackImageReference`, the same helper the manual rollback API uses. Under a `required` post-pull identity policy with no establishable digest it refuses with the rollback-failure audit and runs nothing destructive; otherwise it pins `repo:tag@sha256:...`. Non-required keeps the mutable-tag fallback with a warning.
- DR-65: the Portainer digest pin keeps an explicit `docker.io/` prefix via the compose path's `preserveExplicitDockerIoPrefix`.
- DR-63: `api-key-enforcement.integration.test.ts` builds its app, argon2id authenticator and server once per file instead of once per case; that hash was the fixed cost that pushed cases past 5000 ms under the parallel gate. DR-74 (rate-limit header flake) left alone, no confident root cause.
- CI-11: the demo site sends `Cross-Origin-Opener-Policy: same-origin` (ZAP 90004 checks COOP and COEP together; #878 only added COEP).

App 448 files / 15122 tests, 100% coverage.

## Changelog

- 🐛 Fixed auto-rollback identity binding:
  - Required policies reject rollback when no digest can be established.
  - Rollback failures create error audit events.
  - Successful binding pins the rollback image to its digest.
  - Non-required policies retain mutable-tag fallback with a warning.
- 🔧 Changed `HealthMonitor` to resolve rollback references through identity binding.
- 🐛 Fixed Portainer digest pinning to preserve explicit `docker.io/` prefixes.
- 🔧 Optimized API-key enforcement integration tests by sharing setup per test file.
- 🔒 Added `Cross-Origin-Opener-Policy: same-origin` to the demo site for CI-11 and ZAP compliance.
- ✨ Exported `preserveExplicitDockerIoPrefix` for reuse by Portainer stack pinning.

## Concerns

- Verify all trigger providers implement the optional identity-binding hooks consistently.
- Verify audit events include the expected rollback image and policy context.
- Verify the shared integration-test setup does not leak mutable state between tests.


biggest-littlest (approved):
DR-59, DR-65, DR-63 and the demo COOP header. Gate green, CI green on the refreshed merge.

ALARGECOMPANY (approved):
DR-59, DR-65, DR-63 and the demo COOP header. Gate green, CI green on the refreshed merge.

PR #1056, opened 2026-09-06
build(docker): bump the cosign apk pin to 3.0.6-r2

scttbnsn:
Alpine replaced cosign 3.0.6-r1 with r2, so the Docker build check fails on every open PR and the next release cut would fail the same way. Pin follows. Same change on each maintained line.


biggest-littlest (approved):
Pin follows alpine's cosign 3.0.6-r2; multi-arch smoke build green.

ALARGECOMPANY (approved):
Pin follows alpine's cosign 3.0.6-r2; multi-arch smoke build green.

PR #1061, opened 2026-09-06
fix(v1.8): pinned-tag shape check, rename events, and handshake body validation

scttbnsn:
Three hardening items from the 1.8 roadmap (DR-23, DR-24, DR-25).

- `isTagPinned()` fell through to an unconditional `return true` after its numeric-shape check, so bare distro and variant tags like `redis:alpine` or `node:slim` read as pinned and disappeared under Hide Pinned. The tail now returns the shape check's own result. The watchers doc also named the wrong field for Hide Pinned (`tagPrecision` instead of `tagPinned`).
- A `docker rename` or a display-name change was persisted without a `containerUpdated` event because `hasContainerChangedWithSecurityHashes` compared neither `name` nor `displayName`. SSE rows, Prometheus labels and MQTT topics stayed on the old name until an unrelated field changed. Both fields are now part of the comparison.
- `AgentClient._doHandshake` accepted any `/api/containers` body, so a non-array response threw after `deregisterAgentComponents` had run and before triggers were re-registered, and `dd:ack` fires once per connection so nothing retried. The body is now checked with `Array.isArray` before any registry mutation.

Tests cover each path. Coverage stays at 100%.


biggest-littlest (approved):
DR-23, DR-24, DR-25 hardening with tests for each path. Gate green, CI green.

ALARGECOMPANY (approved):
DR-23, DR-24, DR-25 hardening with tests for each path. Gate green, CI green.

biggest-littlest (approved):
DR-23, DR-24, DR-25 hardening with tests for each path. Rebased onto dev/v1.8, gate green, CI green.

ALARGECOMPANY (approved):
DR-23, DR-24, DR-25 hardening with tests for each path. Rebased onto dev/v1.8, gate green, CI green.

PR #1062, opened 2026-09-06
fix(v1.6): give the session store its own file instead of sharing dd.json

scttbnsn:
1.6 port of DR-121 (#1055 on dev/v1.8).

The connect-loki session store and the main store both wrote `/store/dd.json`, and whichever saved last erased the other's data. express-session touches the session on every authenticated read, so the session store rewrote the file within seconds while the main store autosaved every five minutes. Sessions now live in `dd-sessions.json` next to the main store via `getSessionStorePath()`. A legacy `Sessions` collection left in dd.json is dropped on init and the drop is persisted once.

Three commits cherry-picked with -x. The only adaptation is that 1.6 has no `app/api/session-principal.ts`, so the comment-only hunk there was dropped; the wiring in `app/api/auth.ts` applied cleanly.

## Changelog

- ✨ Added `getSessionStorePath()` to derive a dedicated sibling session database file.
- 🔧 Changed authentication to use the dedicated session store path.
- 🐛 Fixed cross-store saves, session touches, and restarts so session data and application data do not overwrite each other.
- 🗑️ Removed legacy `Sessions` collections from the main database and persisted the cleanup immediately.
- ✨ Added coverage for path derivation, persistence, reloads, and legacy cleanup.
- 🔧 Updated storage documentation and changelog details.


biggest-littlest (approved):
1.6 port of DR-121, three commits cherry-picked with -x, session-principal hunk dropped as the file does not exist on 1.6. Gate green, CI green.

ALARGECOMPANY (approved):
1.6 port of DR-121, three commits cherry-picked with -x, session-principal hunk dropped as the file does not exist on 1.6. Gate green, CI green.

PR #1063, opened 2026-09-06
fix(v1.7): give the session store its own file instead of sharing dd.json

scttbnsn:
1.7 port of DR-121 (#1055 on dev/v1.8).

The connect-loki session store and the main store both wrote `/store/dd.json`, and whichever saved last erased the other's data. express-session touches the session on every authenticated read, so the session store rewrote the file within seconds while the main store autosaved every five minutes. Sessions now live in `dd-sessions.json` next to the main store via `getSessionStorePath()`. A legacy `Sessions` collection left in dd.json is dropped on init and the drop is persisted once.

Three commits cherry-picked with -x. The only adaptation is that 1.7 has no `app/api/session-principal.ts`, so the comment-only hunk there was dropped; the wiring in `app/api/auth.ts` applied cleanly.


biggest-littlest (approved):
1.7 port of DR-121, three commits cherry-picked with -x, session-principal hunk dropped as the file does not exist on 1.7. Gate green, CI green.

ALARGECOMPANY (approved):
1.7 port of DR-121, three commits cherry-picked with -x, session-principal hunk dropped as the file does not exist on 1.7. Gate green, CI green.

PR #1065, opened 2026-09-06
feat(store): open dd.sqlite alongside dd.json and add DD_STORE_DB_FILE

scttbnsn:
7-STORE slice 2: the store opens both engines. `init()` opens the SQLite database from slice 1 next to `dd.json` (`dd.sqlite`, configurable with `DD_STORE_DB_FILE`), runs the migrations, and applies the same file permissions to the database and its `-wal` and `-shm` sidecars. `save()` still writes the LokiJS file and now checkpoints SQLite. The debug snapshot reports both engines, memory mode opens `:memory:` for both, and `GET /api/store` gains an additive `dbFile` field. Nothing reads from SQLite yet.

Also folds in the slice 1 review carry-over: the open pragmas now take the journal mode actually granted, and a writable file-backed database that fell back from WAL to a rollback journal gets `synchronous=FULL` instead of NORMAL.

`DD_STORE_DB_FILE` is read as its own flat field rather than through the dotted-path helper, because its two-word suffix would otherwise nest as `db.file` and fail the store schema. Same pattern as `DD_AGENT_ALLOW_INSECURE_SECRET`.

Tests cover the default and override paths, both files created, permissions on all three files, memory mode, and the additive API response. Coverage stays at 100%.


biggest-littlest (approved):
7-STORE slice 2: dual-open, DD_STORE_DB_FILE, checkpoint on save, synchronous=FULL under a rollback-journal fallback. Rebased over DR-121, gate green, CI green, only Vercel quota red.

ALARGECOMPANY (approved):
7-STORE slice 2: dual-open, DD_STORE_DB_FILE, checkpoint on save, synchronous=FULL under a rollback-journal fallback. Rebased over DR-121, gate green, CI green, only Vercel quota red.

PR #1066, opened 2026-09-06
chore(release): prepare v1.7.0-rc.12

scttbnsn:
Release-prepare commit for v1.7.0-rc.12, same 25-file shape as the rc.11 one.

Moves the six Unreleased Fixed bullets (demo COOP header, arm64 arch-check digest overwrite, the two once-notification fixes and the prune policy stash from the 1.7 port batch, DR-121 session store file) and the agents-page TLS doc fix under `## [1.7.0-rc.12] — 2026-09-06`, adds the README highlight blocks in all seven languages plus the docs updates page entry, and bumps the version string in the demo mocks, site config, API docs and release tests.

### DIFF
--- a/.github/tests/readme-translations.test.ts
+++ b/.github/tests/readme-translations.test.ts
@@ -66,114 +66,123 @@ const localizedSurfaceFragments: Record<
     featureTableHeader: '| | Funktion | Beschreibung |',
     builtWithHeading: '<h2 align="center" id="built-with">Gebaut mit</h2>',
     communityQaHeading: '### Community-QA',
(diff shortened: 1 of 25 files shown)


biggest-littlest (approved):
v1.7.0-rc.12 prepare commit, same 25-file shape as rc.11. Gate green, CI green, only the Vercel daily quota red.

ALARGECOMPANY (approved):
v1.7.0-rc.12 prepare commit, same 25-file shape as rc.11. Gate green, CI green, only the Vercel daily quota red.

PR #1067, opened 2026-09-06
fix(v1.8): keep dd.watch.digest across Portwing reports and share the lookup-label helper

scttbnsn:
Two hardening items from the 1.8 roadmap (DR-38, DR-39).

- `dd.watch.digest=true` was dead for controller-Docker-transport (Portwing) containers. Portwing hardcodes `image.digest.watch: false` on every runtime-state report, and `preserveControllerDockerEnrichment` restored a fixed list of controller-owned fields that did not include `image`, so the label-derived value was lost on the first report after discovery. The stored `image.digest.watch` is now restored alongside the other controller-owned fields. A test reproduced the loss before the fix; a companion test checks an unlabelled container still ends up with `digest.watch: false`.
- `applyRegistryLookupLabels` duplicated the unexported `resolveLookupImageFromContainerLabels`. The helper is now exported and the agent path calls it, keeping its existing precedence where an agent-reported lookup image beats the label. Refactor only, no changelog bullet.

Coverage stays at 100%.

## Changelog

- ✨ Added shared `resolveLookupImageFromContainerLabels` export for container initialization and agent lookup.
- 🐛 Fixed controller-Docker runtime-state enrichment to preserve `image.digest.watch` from stored container state.
- 🔧 Preserved agent-reported lookup image precedence over label-derived values.
- ✨ Added tests for digest-watch restoration, missing `image`, and unlabelled containers.

## Concerns

- Review whether the new public export follows the provider module export convention.


biggest-littlest (approved):
DR-38 with a reproducing test, DR-39 helper shared. Gate green, CI green, only the Vercel daily quota red.

ALARGECOMPANY (approved):
DR-38 with a reproducing test, DR-39 helper shared. Gate green, CI green, only the Vercel daily quota red.

PR #1069, opened 2026-09-06
chore(release): promote v1.7.0-rc.12 to main

scttbnsn:
Promotes dev/v1.7 to main for the v1.7.0-rc.12 cut.

Since rc.11: the 1.7 port batch (#1050), the cosign apk pin bump (#1057), the gitleaks ignore entries for the shifted webhook doc lines (#1054), DR-121 session store file (#1063), and the rc.12 prepare commit. Tree equality with dev/v1.7 is asserted after merge and by the release-cut drift check.

PR #1054, opened 2026-09-06
chore(security): ignore the shifted webhook curl examples for the 1.7 port batch

scttbnsn:
Additive gitleaks ignore entries so #1050 can pass the Secrets job, which scans the tracked tree under the base branch's ignore file. Same shape as #1043 on dev/v1.8. Lines 48/117/124/131 of the webhooks docs page are the existing curl examples moved down by the DR-60 docs edit.

🔒 Security
- Added four `.gitleaksignore` entries for shifted `curl-auth-header` findings in the webhooks documentation.

PR #1071, opened 2026-09-06
chore(security): carry the webhook curl-example ignore entries on main for the rc.12 promotion

scttbnsn:
Policy file only. The Secrets job scans a PR with the base branch's `.gitleaksignore`, and the rc.12 promotion (#1069) fails because main's copy predates the webhook doc line shift that #1054 covered on dev/v1.7. This adds the same four line-pinned entries to main, byte-identical to dev/v1.7's file, so the promotion's tree still equals dev/v1.7 after both land. main stays untagged only until the rc.12 cut tags it right after.

### Changelog

- 🔒 Security: Added four line-pinned `gitleaks` ignore entries for `curl-auth-header` detections in `content/docs/current/configuration/webhooks/index.mdx`.
- 🐛 Fixed: Updated `.gitleaksignore` to match `dev/v1.7` and prevent Secrets job failures during rc.12 promotion.

### Concerns

- Confirm the entries remain valid if webhook documentation line numbers change.
- Verify tree parity with `dev/v1.7` after both changes land.

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -341,6 +341,10 @@ content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:115
 content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:122
 content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:129
 content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:46
+content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:48
+content/docs/current/configuration/webhooks/index.mdx:curl-auth-header:117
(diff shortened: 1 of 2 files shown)


biggest-littlest (approved):
Policy-only change to main so the rc.12 promotion's Secrets job sees the current webhook doc lines, plus the cosign r2 pin main's Build needs. Both files are byte-identical to dev/v1.7.

ALARGECOMPANY (approved):
Policy-only change to main so the rc.12 promotion's Secrets job sees the current webhook doc lines, plus the cosign r2 pin main's Build needs. Both files are byte-identical to dev/v1.7.

PR #1073, opened 2026-09-06
feat(store): move app, secrets, settings and ui-preferences onto SQLite

scttbnsn:
7-STORE slice 3: the leaf singletons move over. `app` (version record), `secrets` (session secret), `settings` and `ui-preferences` now read and write SQLite instead of their LokiJS collections. Each module keeps its public function signatures, so nothing above the store changes.

First start on an existing install imports those four collections out of `dd.json` before anything reads them, using the collection importers added in `app/store/db/importers/`. The import is idempotent and skipped once the SQLite side already holds the data. A `dd-v1.7.json` fixture exercises the real import path end to end.

`DD_STORE_DB_FILE` docs and the changelog bullet now say what actually lives in the database. Everything else (containers, audit, backups, agents) still lives in `DD_STORE_FILE` and moves in later slices.

Tests cover the four modules against SQLite, the importers, the first-start import and the skip path. Coverage stays at 100%.


biggest-littlest (approved):
7-STORE slice 3: app, secrets, settings and ui-preferences on SQLite with the first-start import. Gate and CI green; qlty is the out-of-minutes noise.

ALARGECOMPANY (approved):
7-STORE slice 3: app, secrets, settings and ui-preferences on SQLite with the first-start import. Gate and CI green; qlty is the out-of-minutes noise.

PR #1074, opened 2026-09-07
feat(store): move agent keys, name bindings and API keys onto SQLite

scttbnsn:
7-STORE slice 4: the identity and credential tables move over. `agent-keys`, `name-bindings` and `api-key` now read and write SQLite with their public function signatures unchanged, so the API, agent and portwing hello paths don't change. API key scopes become rows in the `api_key_scope` join table the schema already promised, and the record still exposes `scopes: string[]` in the original order.

Cascade revocation runs as the same breadth-first walk inside one transaction, with a test that pins the `revokedKeyIds` order against the LokiJS implementation for a three-level tree with siblings.

First start on an existing install imports the three collections out of `dd.json` through new importers. The `dd-v1.7.json` fixture gains an API key with two scopes and a child key, an agent key and a name binding, and the first-start import test asserts a `ddk_` credential minted under v1.7 still verifies, `deleteBindingsForKey` releases the same names, and the imported agent key is served by `getKey` and `listKeys`.

The three API tests that built a real LokiJS instance now run against the migrated in-memory database. Changelog bullet and the `DD_STORE_DB_FILE` doc row extended, no new bullet.

Focused suites: 20 files, 420 tests green. Coverage stays at 100%.

## Changelog

- ✨ Added SQLite importers for `agent-keys`, `name-bindings`, and `api-keys`.
- ✨ Added migration coverage for scoped and child API keys, agent keys, and name bindings.
- 🔧 Changed the three stores from LokiJS collections to SQLite tables.
- 🔧 Preserved store function signatures and return shapes.
- 🔧 Stored API key scopes in `api_key_scope` while preserving insertion order.
- 🔧 Kept breadth-first API key cascade revocation transactional and order-preserving.
- 🔧 Migrated API and Portwing tests to in-memory SQLite.
- 🔧 Updated SQLite storage documentation and fixtures.

## Concerns

- Verify all production callers pass the shared SQLite `Database` to the updated `createCollections` functions.
- Verify foreign-key cleanup removes `api_key_scope` rows when API keys are pruned or deleted.
- Verify first-start import remains idempotent when `dd.json` contains invalid or duplicate records.
- Verify SQLite database handles are reset after test teardown and failed initialization.


biggest-littlest (approved):
7-STORE slice 4: agent keys, name bindings and API keys on SQLite with importers and the first-start import. Gate and CI green; qlty is the out-of-minutes noise.

ALARGECOMPANY (approved):
7-STORE slice 4: agent keys, name bindings and API keys on SQLite with importers and the first-start import. Gate and CI green; qlty is the out-of-minutes noise.

PR #1075, opened 2026-09-07
fix(v1.8): platform check in compose preflight, per-replica binding policy, digest exclusion log

scttbnsn:
Three roadmap items, one commit each.

DR-41: the compose-file-once preflight covered signature, scan and SBOM but skipped the platform compatibility check, which only ran inside the per-service refresh. A single-arch image that pulled cleanly but was wrong for the host failed after earlier services were already recreated, leaving a partial apply and a partially restored file. The pure inspect now runs in `buildComposeFileOnceRuntimeContextByService` right after `capturePulledImageIdentity`, against the bound identity, so the whole file is rejected before anything is touched.

DR-42: identity-binding policy was captured once per service from whichever replica resolved first and reused for every replica, so two hand-run replicas mapped to the same service with different `dd.security.gate` labels got the wrong outcome in both directions, and the unbound warning audit row fired twice per first replica. Capture now stores the raw identity or failure reason per service, the policy is applied per container inside `runComposeFileOncePostPullGate`, and exactly one audit row lands per replica.

DR-95: the single-container digest path dropped a container silently when `mustTrigger()` said no. It now logs the same debug line the batch path does, naming the container and the reason.

Tests cover each path (preflight platform rejection with nothing recreated, two replicas with differing gate labels under a bind failure, the digest debug line). Changelog bullets under Fixed. Coverage stays at 100%.

### 🐛 Fixed

- Reject incompatible pulled-image platforms during compose-file-once preflight.
- Apply identity-binding policy per replica.
- Record binding failures according to each replica’s `dd.security.gate` label.
- Create one skipped-scan audit row per applicable replica.
- Log the container and exclusion reason when `mustTrigger()` rejects a digest report.
- Prevent rejected containers from entering the digest buffer.

### Tests

- Added coverage for platform rejection before recreation or compose-file rewrite.
- Added coverage for per-replica gate handling during bind failure.
- Added coverage for digest exclusion logging.
- Maintained 100% coverage.


biggest-littlest (approved):
DR-41, DR-42 and DR-95: preflight platform check, per-replica binding policy, digest exclusion log line. Gate and CI green; qlty is the out-of-minutes noise.

ALARGECOMPANY (approved):
DR-41, DR-42 and DR-95: preflight platform check, per-replica binding policy, digest exclusion log line. Gate and CI green; qlty is the out-of-minutes noise.

PR #1077, opened 2026-09-07
ci: track the Dockerfile apk pins and allowlist doc placeholder credentials by pattern

scttbnsn:
Two CI roadmap items, one commit each.

CI-15: nothing watched the Alpine apk pins in the Dockerfile, so when upstream rotated cosign 3.0.6-r1 to r2 on 2026-09-06 every open PR on every line went red at once and the next release cut would have failed the same way. A Renovate custom regex manager now extracts every `name=version-rN` pin from the `apk add` lines and resolves them through the repology datasource against `alpine_3_24`, which is the Alpine release the base image uses. A workflow test runs the manager's regex against the live Dockerfile and asserts all ten pins match, including the cosign line. Whether repology carries an entry for every one of these packages is only provable by the first real Renovate run; the two less common names are `su-exec` and `tini`.

CI-16: the Secrets job checks out the base branch's `.gitleaksignore`, and a promotion's base is `main`, which only gains ignore entries through the promotion itself, so the line-pinned fingerprints for the webhook curl examples failed the rc.12 promotion after the docs shifted two lines. The documented placeholder value is now allowlisted by pattern, scoped to `content/docs/` with `condition = "and"`, and the 47 line-pinned fingerprints under `content/docs/` are removed. The existing workflow test that banned any `paths =` entry in `.gitleaks.toml` is narrowed to allow a path allowlist only when it is paired with a regex match, so a blanket directory exemption still fails it. A new test writes a shifted copy of the webhooks page to a temp dir and asserts zero findings, and a control asserts a real-looking credential outside `content/docs/` still fires.

Workflow tests: 28 files, 266 tests green. Local gitleaks 8.30.1 (the pinned version) finds nothing under `content/docs/`.


biggest-littlest (approved):
CI-15 and CI-16: Renovate tracks the Dockerfile apk pins, doc placeholder credentials are allowlisted by pattern. Gate and CI green; Vercel and qlty failures are the daily quota and out-of-minutes noise.

ALARGECOMPANY (approved):
CI-15 and CI-16: Renovate tracks the Dockerfile apk pins, doc placeholder credentials are allowlisted by pattern. Gate and CI green; Vercel and qlty failures are the daily quota and out-of-minutes noise.

PR #1079, opened 2026-09-07
fix(dockercompose): keep the post-pull gate outcome off the caller-supplied context

scttbnsn:
DR-52. `trigger()` and `triggerBatch()` take an `unknown` runtime context and threaded it through to the compose refresh unsanitised, and the internal preflight context was spread after the caller's, so a caller-supplied `imageIdentity`, `securityGateUnboundWarn`, `securityGateUnboundReason` or `securityGateUnboundWarnRecorded` survived whenever compose-file-once was off or a service had not been preflighted. Those keys decide which image the post-pull gate verifies and whether it runs at all. Not externally reachable today (the agent trigger API whitelists its context to operation ids and every other caller is internal), which is why the earlier CWE-693 report was refuted, but the shape invited the bug.

The four keys now live in an internal-only outcome type and are stripped from the caller's argument once at the `triggerBatch()` entry. A test drives the real compose refresh through `trigger()` with a forged context and asserts the gate runs against the identity captured internally. It fails without the fix.

Coverage stays at 100%.


biggest-littlest (approved):
DR-52: post-pull gate outcome moved to an internal-only type and stripped from the caller-supplied context. Gate and CI green; Vercel and qlty failures are the daily quota and out-of-minutes noise.

ALARGECOMPANY (approved):
DR-52: post-pull gate outcome moved to an internal-only type and stripped from the caller-supplied context. Gate and CI green; Vercel and qlty failures are the daily quota and out-of-minutes noise.

PR #1080, opened 2026-09-07
ci(dast): bound the ZAP spider and active scan inside the job budget

scttbnsn:
CI-14. Both weekly `security-dast-web.yml` runs to date (32948622205, 33630040351) had the `DAST: getdrydock.com (ZAP)` job cancelled at its 60-minute timeout. Unbounded, the spider plus the full active scan against the docs site outgrows the budget, no report gets written, and the gate never fires. The step logs for the cancelled scan aren't retained, so the split between spider and active scan isn't measurable from the runs.

The scan step now passes `cmd_options` capping the spider at 10 minutes and the active scan at 35 minutes overall and 5 per rule (ZAP's `scanner.maxScanDurationInMins` and `scanner.maxRuleDurationInMins`), leaving about 15 minutes for startup, the passive scan and the report. A workflow test parses the options and asserts spider plus scan fits inside the job timeout with that headroom, so lowering the timeout or raising the caps out of proportion fails the test.

This reaches `main` with the GA promotion, where the weekly cron runs; the next weekly run after that is the proof. Same change lands on dev/v1.8 in a sibling PR.

## Changelog

- 🔧 Changed ZAP DAST limits:
  - Spider: 10 minutes.
  - Active scan: 35 minutes.
  - Per-rule scan: 5 minutes.
- ✨ Added a regression test for scan limits and the 60-minute job budget.
- 🔒 Prevented ZAP jobs from timing out before report generation and gate evaluation.

## Concerns

- Verify the test parses all `cmd_options` values used by the workflow.
- Verify the 15-minute headroom covers startup, passive scanning, and report generation.

### DIFF
--- a/.github/tests/security-dast-web-workflow.test.ts
+++ b/.github/tests/security-dast-web-workflow.test.ts
@@ -89,9 +89,13 @@ test('the ZAP full scan bounds its spider and active scan inside the job budget'
   const scanMinutes = Number(cmdOptions.match(/scanner\.maxScanDurationInMins=(\d+)/u)?.[1]);
   const ruleMinutes = Number(cmdOptions.match(/scanner\.maxRuleDurationInMins=(\d+)/u)?.[1]);
 
-  expect(spiderMinutes).toBeGreaterThan(0);
-  expect(scanMinutes).toBeGreaterThan(0);
-  expect(ruleMinutes).toBeGreaterThan(0);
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
ci(dast): bound the ZAP spider and active scan inside the job budget

ALARGECOMPANY (approved):
ci(dast): bound the ZAP spider and active scan inside the job budget

PR #1086, opened 2026-09-07
feat(store): move notification rules and approvals onto SQLite

scttbnsn:
7-STORE slice 6: notification rules and the approval queue move over. `app/store/notification.ts` and `app/store/approval.ts` now read and write SQLite with their public function signatures unchanged. The rules table keeps the seeded defaults and the bell thresholds; the approvals table keeps the 30-day decided-row retention prune. `normalizeRules` drops its dead defensive branches now that every caller hands it a typed array.

First start on an existing install imports both collections out of `dd.json` through two new idempotent importers registered in the importer index. The `dd-v1.7.json` fixture gains a rules row plus a pending and a decided approval, and the first-start import test asserts the rules and the queue after import. The decided fixture row is patched to a recent timestamp in the test so the startup prune does not eat it.

`app/api/notification.test.ts` and `app/api/approvals.test.ts` needed no change (they mock the store modules). `store/index.test.ts` now keeps the real notification exports and overrides only `createCollections`, because the importer registry imports the module for real. The reconcile test moves off its LokiJS fake onto the migrated in-memory database.

Focused suites: 14 files, 1129 tests green. gitleaks over the branch: no leaks. Coverage stays at 100%.


biggest-littlest (approved):
feat(store): move notification rules and approvals onto SQLite

ALARGECOMPANY (approved):
feat(store): move notification rules and approvals onto SQLite

PR #1101, opened 2026-09-07
fix(updates): accept a replacement container's in-progress operation without a persisted new id

scttbnsn:
DR-122, the pass-two follow-up from #1099. When an update never persisted `newContainerId`, the replacement container was found by identity in `reconcileInProgressContainerUpdateOperation` and then discarded by the container-id cross-check, so its operation sat in progress until expiry. The by-name fallback before slice 10 behaved the same way.

The cross-check stays, since two containers with the same identity key and different ids must not share an operation. A same-identity, different-id fallback match is now accepted only when the operation's persisted `targetImage` matches the current container's image name and tag, compared raw and through the existing `withoutImageDigest()` helper that the by-name path already used for the same ambiguity. A mismatch is still rejected. Three tests: the accepted replacement, the rejected collision, and the rejected match with no image data; the existing `new_container_id` primary-match test is unchanged.

`triggers/providers/docker` 45 files and 1872 tests green, `ContainerUpdateExecutor.ts` at 100% on every metric.


biggest-littlest (approved):
DR-122

ALARGECOMPANY (approved):
DR-122

Issue #669, opened 2026-08-07
getRepoDigest() trusts RepoDigests[0], anchoring digest comparison to a possibly-stale manifest


scttbnsn:
Found by an adversarially-verified audit of the digest-update pipeline, prompted by discussion #657 (digest updates reported on fully-pinned tags). Present in every version since digest watching shipped; byte-identical in v1.5.2 and the v1.6 GA candidate. Not a regression and not GA-gating.

## The defect

`getRepoDigest()` (`app/watchers/providers/docker/docker-helpers.ts:403-410`) unconditionally returns `RepoDigests[0].split("@")[1]` with no check that the entry belongs to the image reference the container is actually running.

A local Docker image object can legitimately carry multiple `repo@digest` pairs for one Image ID (pull/retag cycles that share cached layers accumulate entries, with no guaranteed ordering that keeps the current one first). When index 0 is a stale or foreign entry, everything downstream anchors to the wrong manifest:

- `container.image.digest.repo` is set from it at initial discovery (`docker-image-details-orchestration.ts:613`) and re-derived from the same index on **every refresh cycle** (`:325`, written at `:397-404`) — so the bad pick self-reinforces instead of self-correcting.
- `handleDigestWatch` (`image-comparison.ts:242-277`) then uses `digest.repo` as the lookup key for the normalizing second `getImageManifestDigest` call and writes the result straight into `container.image.digest.value`, with no cross-check against what the container's own tag currently resolves to.
- `getRawDigestUpdate`/`hasRawUpdate` (`app/model/container.ts:582-661`) compare that value against the tag-derived `result.digest` and report `updateKind: digest`.

## Symptom

A **persistent** digest-update false positive: flagged every poll, and it survives applying the update, because the registry still serves the stale manifest digest and nothing re-derives `digest.repo` from anything but `RepoDigests[0]`. Distinguishable from a genuine same-tag republish precisely by that immortality — a real republish clears after one apply.

## What the audit ruled out (so nobody re-chases these)

- The `tagsCandidates[0]` substitution in `handleDigestWatch` cannot fire for pinned tags — candidates are always empty for them under default settings.
- No manifest-list vs platform-digest mismatch class: Accept headers and platform filtering in `Registry.ts:254-368` are correct, and the poll-cycle digest cache keys (`BaseRegistry.ts:124-149`) don't cross-contaminate.
[...]

PR #1108, opened 2026-09-07
fix(watcher): re-derive digest watching on every scan so stale rows flag updates

### LINKED ISSUE #1070: Some portainer_agent containers not showing updates
### Drydock version

v1.7.0-rc.11

### What happened?

See attached screenshot

Image

I verified manually that all portainer containers are running 2.39.  I Tried forcing an update check on the missing agent containers but no luck.

### What did you expect?

See above

### Steps to reproduce

See above

### Relevant configuration

```yaml

```

### Logs

```text

```

### Installation method

Docker Compose

### Docker version

24.0.1

### OS / Architecture

Synology DSM 7

scttbnsn:
Fixes #1070.

`image.digest.watch` was written once at discovery and only rewritten by the unresolved-tag repair path. The default for a non-semver Docker Hub tag flipped from off to on in v1.5.0-rc.17, so any `something:latest` container drydock had been watching since before May still carried `watch: false`, `findNewVersion` skipped the digest block, and the row said Current no matter what the registry served. Containers recreated after that release got a fresh row and worked, which is why the reporter saw two of five `portainer/agent:latest` flag the 2.45.0 digest and two stay Current.

`refreshStoredContainerImageFields` now runs `resolveContainerImageState` every cycle and writes the derived `watchDigest` back, the same way `isLocalImage` and `repoDigests` are already refreshed for #669. An explicit `dd.watch.digest` label or imgset value is read from the live container each cycle, so an operator's choice still wins; only the stale auto-derived default is replaced. Three tests cover the stale-false row, the label override on a floating tag, and the label override on a semver tag. CHANGELOG entry included.

Forward-port to dev/v1.8 and port to dev/v1.6 follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
DR-124 digest watch re-derive

ALARGECOMPANY (approved):
DR-124 digest watch re-derive

PR #1111, opened 2026-09-07
fix(watcher): re-derive digest watching on every scan so stale rows flag updates

### LINKED ISSUE #1070: Some portainer_agent containers not showing updates

scttbnsn:
Forward-port of #1108 from dev/v1.7. Fixes #1070.

`image.digest.watch` was written once at discovery and only rewritten by the unresolved-tag repair path, so any `something:latest` container on Docker Hub that drydock had been watching since before v1.5.0-rc.17 (when the default flipped on) still carried `watch: false` and reported Current no matter what the registry served. `refreshStoredContainerImageFields` now re-derives it every cycle from the live container; an explicit `dd.watch.digest` label or imgset value still wins. Applied clean apart from the CHANGELOG placement.


biggest-littlest (approved):
DR-124 1.8 forward-port

ALARGECOMPANY (approved):
DR-124 1.8 forward-port

PR #1112, opened 2026-09-07
feat(api): expose the effective configuration and validate candidates without applying them

scttbnsn:
Roadmap 7.1 slices 4 and 5, on top of #1106. Spec: `.planning/roadmap/spec-7.1-config-file.md` sections 4.1, 4.2 and the slice notes.

Slice 4, `feat(api): expose the effective configuration`: `GET /api/v1/config` and `GET /api/v1/config/:section` return the effective tree (env over file over defaults) with a per-key `source` of `env` or `file`, redacted uniformly through `redactConfigurationTree`, session-only scope (decision D4), 5 requests per 60 s. Sections come from the `DD_` prefixes the loader already knows, so a new provider shows up without a route change. The loader records the file it read (`getConfigFileInfo`), which the response carries as `file.path` and `file.modifiedAt`. Audit `config-read`.

Slice 5, `feat(api): validate a candidate configuration without applying it`: `POST /api/v1/config/validate` takes `{ yaml }` or a tree, admin scope, and always answers 200 with `{ valid, errors, diff: { changed, reload, restart } }`, the diff computed against the running configuration and classified by which keys can reload and which need a restart. Nothing is applied. Audit `config-validated`. Integration test drives it end to end through the real loader and Joi schemas.

One rebase conflict against the #1106 squash (the loader's single-handle read) resolved by capturing the file info from the same `handle.stat()`.


biggest-littlest (approved):
7.1 slices 4+5

ALARGECOMPANY (approved):
7.1 slices 4+5

PR #1114, opened 2026-09-07
docs(tags): document the tag transform grammar

scttbnsn:
Roadmap DOC-5, first item. The watchers page gains a Grammar subsection under "Transform the tags": RE2 flavour (no lookaround or backreferences, `(?i)` for case-insensitive), first `=>` wins, one pattern and replacement pair, passthrough when there is no `=>` or no match, `$1..$n` placeholders with a missing group becoming empty, the 1024-character pattern cap. An eight-row worked-example table, each row run through the real `transformTag`, plus a callout showing what a compile failure looks like.

A second subsection covers the interaction with include, exclude and semver comparison: include and exclude run on raw registry tags, the transform is applied after filtering and its output is what gets parsed, compared and classified, and `${transformed}` in `dd.link.template` is that same value.

The existing "validated at startup" callout was wrong. The code path is a per-container catch in the Docker watcher that logs `Failed to fetch image detail` every scan and never crashes the process, so the callout now says that.


biggest-littlest (approved):
DOC-5 tag transform grammar

ALARGECOMPANY (approved):
DOC-5 tag transform grammar

PR #1117, opened 2026-09-07
docs(watchers): document dd.link.template and the link placeholders

scttbnsn:
Roadmap DOC-5, second item. The "Associate a link to the container version" section on the watchers page now covers the label and the imgset form (`DD_WATCHER_{watcher}_IMGSET_{imgset}_LINK_TEMPLATE`) with label-over-imgset precedence, the full placeholder table (`${original}`, the deprecated `${raw}` alias, `${transformed}`, and the semver-gated `${major}`, `${minor}`, `${patch}`, `${prerelease}`), three worked examples run through the real `getLink`, and the failure behaviour: an unknown placeholder becomes an empty string with no error, and a value that is not an `http(s)` URL is stored but the UI's release-link renderer drops it. The labels table row and the tag transform grammar section link to it.


biggest-littlest (approved):
DOC-5 item 2

ALARGECOMPANY (approved):
DOC-5 item 2

PR #1120, opened 2026-09-07
build(deps): forward-port the v1.7 Renovate batches to dev/v1.8

scttbnsn:
Roadmap DEP-2. Renovate targets `dev/v1.7` only, and the batches that landed there since the v1.8 cut were never carried over: #1083 (root), #1084 (apps/demo), #1082 (github-actions), #1085 and #1098 (apps/web), #1087 (ui), #1088 (e2e), #1089 (app). This cherry-picks those eight commits in order with `-x` trailers. Every one applied cleanly, the regenerated lockfiles were byte-identical to what the cherry-picks already carried, and the pin-assertion tests (biome `$schema`, analytics-integration, security-dependency-versions, picomatch lockfile, harden-runner and crowdin workflow tests) already match. The 1.8-only removals (`lokijs`, `connect-loki`, `passport*`, `@types/passport`) stay removed; the 1.8-only workflow additions (CI-17 memory budget, PR-scoped secrets scan, scorecard permissions) are preserved around the new action pins. After this every `package.json` and every `uses:` pin equals `dev/v1.7`'s.

Found while forward-porting #1098: it would not cherry-pick because `dev/v1.8` still pinned posthog-js 1.417.1.


biggest-littlest (approved):
DEP-2 deps forward-port

ALARGECOMPANY (approved):
DEP-2 deps forward-port

PR #1123, opened 2026-09-07
test(api): make the integration tests deterministic under load

scttbnsn:
Roadmap DR-57 and DR-74. Both flaked only when two pre-push gates ran at once, and the cause is not the wall clock.

`server.listen(0)` with no host binds the IPv6 wildcard. libuv sets `SO_REUSEADDR` on every listener, and on Darwin the ephemeral-port picker then rejects only an exact address-plus-port match, so a wildcard listener can be handed a port another test already holds on `127.0.0.1` (the healthcheck binary test and the Portwing bridge both bind loopback). The kernel routes `127.0.0.1:P` to the more specific socket, so the test's request answers from the other test's server: `expected 404 to be 200` in the session-persistence test, and an empty RateLimit header map in the api-key test (the response never went through the limiter app). Measured on a 14-core Mac: an explicit loopback bind over a live wildcard listener succeeds and wins the connection.

Fix: every test listener binds `listen(0, '127.0.0.1')`. The two flaking files first, then the same one-token change at the other twelve `listen(0` sites in app tests (wudcard, config-validate, api/index, registry webhook, Oidc). The rate-limit case also had a second, latent defect: `RateLimit-Reset` drifts by a second between two independently timed limiters under contention, so the probes now run with `Date` frozen and the reset value is asserted, with `ratelimit-remaining` still doing the discriminating work.

Verified: each fixed file 5 of 5 alone, and `npx vitest run api/` 3 of 3 under 14 CPU burners with a loop churning loopback ports and a full coverage gate running in another worktree. No production code changed. The 1.7 and 1.6 lines get the same change as a port.

### DIFF
--- a/app/api/api-key-enforcement.integration.test.ts
+++ b/app/api/api-key-enforcement.integration.test.ts
@@ -202,10 +202,25 @@ function createTestApp(store: SessionStore): Application {
   return app;
 }
 
+/**
+ * Bind the loopback address, never the wildcard.
+ *
+ * `listen(0)` binds `::` dual-stack, and libuv sets SO_REUSEADDR, which makes
+ * the kernel's ephemeral-port picker check for an *exact* address+port conflict
+ * only. A wildcard listener can be handed a port another process already holds
+ * on `127.0.0.1` (`healthcheck.binary.test.ts` and
+ * `agent/PortwingDockerBridge.ts` both bind that address), and the more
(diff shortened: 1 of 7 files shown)


biggest-littlest (approved):
DR-57 DR-74

ALARGECOMPANY (approved):
DR-57 DR-74

PR #1124, opened 2026-09-07
test(api): make the integration tests deterministic under load

scttbnsn:
Port of #1123 to dev/v1.7.

Every test listener now binds 127.0.0.1 instead of the wildcard address. On Darwin, `listen(0)` on `::` can be handed a port another test already holds on 127.0.0.1 (libuv sets SO_REUSEADDR and the kernel's port check only rejects exact matches), so requests hit a foreign server and the session-persistence assertions fail under load with a 404 and no RateLimit headers.

The api-key-enforcement and config-validate suites do not exist on this line, so those hunks are left out. Roadmap DR-57.

### DIFF
--- a/app/api/auth-session-persistence.integration.test.ts
+++ b/app/api/auth-session-persistence.integration.test.ts
@@ -137,10 +137,26 @@ function createTestApp(store: LokiSessionStore): Application {
   return app;
 }
 
+/**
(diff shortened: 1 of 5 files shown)


biggest-littlest (approved):
DR-57 1.7 port

ALARGECOMPANY (approved):
DR-57 1.7 port

PR #1125, opened 2026-09-07
test(api): bind every test listener to loopback

scttbnsn:
Port of #1123 to dev/v1.6.

Every test listener now binds 127.0.0.1 instead of the wildcard address. On Darwin, `listen(0)` on `::` can be handed a port another test already holds on 127.0.0.1 (libuv sets SO_REUSEADDR and the kernel's port check only rejects exact matches), so requests hit a foreign server and assertions fail under load.

This line has no session-persistence suite, so only the listener binds are ported. Roadmap DR-57.


biggest-littlest (approved):
DR-57 1.6 port

ALARGECOMPANY (approved):
DR-57 1.6 port

PR #1126, opened 2026-09-07
ci(release): label the image with the cut's source commit

scttbnsn:
Roadmap DR-119.

A maintenance cut checks out SOURCE_SHA for the build, but both docker/metadata-action steps left `org.opencontainers.image.revision` at its default of `github.sha`, so the shipped image was labeled with main's head instead of the dev-branch commit it was built from. The v1.6.1-rc.9 staging image carried `5ae315227` where `2969675ef` belonged.

Both steps now set the label explicitly from `steps.source.outputs.source_sha`. A workflow test pins that. Prerelease cuts are unaffected since SOURCE_SHA equals the checkout there.

## Changelog

- 🔧 Set the revision label for published and staging images to `steps.source.outputs.source_sha`.
- ✨ Added workflow tests for the revision label.

### DIFF
--- a/.github/tests/release-cut-retry-workflow.test.ts
+++ b/.github/tests/release-cut-retry-workflow.test.ts
@@ -158,6 +158,26 @@ test('release-cut delegates image tags and labels to docker metadata-action', ()
   expect(shellTagComputations).toStrictEqual([]);
 });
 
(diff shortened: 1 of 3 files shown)


biggest-littlest (approved):
DR-119

ALARGECOMPANY (approved):
DR-119

PR #1132, opened 2026-09-08
fix(docker): sanitize inherited runtime config when recreating a container for rollback

scttbnsn:
Roadmap DR-126, found by the DR-93 audit verification.

`Docker.recreateContainer` passed the log container where the update path passes runtime-config options, so on the auto-rollback (HealthMonitor) and manual restore (backup API) paths the source image was unknown and the UNKNOWN-origin sanitization never fired. An entrypoint or command the newer image introduced was cloned verbatim onto the older image, which lacks the file, and the rolled-back container failed to start.

- `recreateContainer` now resolves clone options through `getCloneRuntimeConfigOptions` with the current spec's image as source and the rollback image as target, the same shape the update path uses.
- Tests: an Entrypoint materialized from the pre-rollback image is dropped when the target image does not define it; a field the target image also defines is preserved.
- CHANGELOG Fixed entry.

Port to v1.7 follows.

## Changelog

- 🐛 Fixed rollback container recreation by sanitizing inherited runtime configuration.
- 🔧 Updated `Docker.recreateContainer` to use `getCloneRuntimeConfigOptions`.
- ✨ Added regression tests for removing stale `Entrypoint` values and preserving target-image values.
- ✨ Added a changelog entry.

## Concerns

- Add regression coverage for stale `Command` values.
- Port the fix to v1.7.


biggest-littlest (approved):
fix(docker): sanitize inherited runtime config when recreating a container for rollback (#1132)

ALARGECOMPANY (approved):
fix(docker): sanitize inherited runtime config when recreating a container for rollback (#1132)

PR #1133, opened 2026-09-08
fix(watcher): require an exact variant suffix match when climbing tag families

scttbnsn:
Roadmap DR-125, found by the DR-93 audit verification.

`isSuffixCompatible` accepted a candidate whose suffix template was a prefix or an extension of the reference's, so under `dd.tag.family=loose` (or with a `dd.tag.include` filter) a container on `1.27.3-alpine` could be offered `1.28.0-alpine-perl`, and `1.27.3-alpine-slim` could climb to `1.28.0-alpine`. Strict mode with a specific tag precision was protected by the pin gate, so default installs were unaffected.

- Suffix templates must now match exactly. The one carve-out is precision growth inside the same variant (`-alpine` to `-alpine3.21`), which only adds digit placeholders and dots, never new letters; that documented case keeps working.
- Tests: an nginx-style variant matrix (bare, `-alpine`, `-alpine-slim`, `-alpine-perl`, `-perl`) under strict and loose, plus a `dd.tag.include=.*` case showing the filter does not reopen the cross-variant pick.
- CHANGELOG Fixed entry. The watcher docs already describe exact-suffix behaviour, so no doc change.

Port to v1.7 follows.

## Changelog

- 🐛 Fixed `isSuffixCompatible` to require exact variant suffix matches.
- ✨ Added support for digit-and-dot precision extensions, such as `-alpine` to `-alpine3.21`.
- ✨ Added tests for strict and loose family matching, include filters, and nginx-style variants.
- ✨ Added a CHANGELOG entry.


biggest-littlest (approved):
fix(watcher): require an exact variant suffix match when climbing tag families (#1133)

ALARGECOMPANY (approved):
fix(watcher): require an exact variant suffix match when climbing tag families (#1133)

biggest-littlest (approved):
fix(watcher): require an exact variant suffix match when climbing tag families (#1133)

ALARGECOMPANY (approved):
fix(watcher): require an exact variant suffix match when climbing tag families (#1133)

PR #1136, opened 2026-09-08
fix(watcher): require a digit placeholder in precision-only suffix extensions

scttbnsn:
Follow-up to #1133. The precision-only carve-out in `isPrecisionOnlyExtension` accepted any extension made of `#` and `.`, so a punctuation-only extension like the trailing dot in `1.3.0-alpine.` counted as the same family as `1.3.0-alpine`. The extension now has to match `^\.?#(?:\.#)*$`: at least one digit placeholder, dots only as separators.

Raised by CodeRabbit on the v1.7 port (#1134). Lands here first, then gets cherry-picked into #1134.

- Rejection tests for `1.3.0-alpine.` and `1.3.0-alpine..`, `1.3.0-alpine.1` still accepted.
- CHANGELOG bullet under Unreleased / Fixed.

## Changelog

- 🐛 Fixed `isPrecisionOnlyExtension` to require at least one digit placeholder.
- 🐛 Rejected punctuation-only Docker tag suffixes such as `.` and `..`.
- ✨ Preserved support for valid numeric extensions such as `.1`.
- ✨ Added regression tests for accepted and rejected suffixes.
- 📝 Added an Unreleased / Fixed changelog entry.


biggest-littlest (approved):
Reviewed the carve-out regex and the rejection tests. Gate and CI green.

ALARGECOMPANY (approved):
Reviewed the carve-out regex and the rejection tests. Gate and CI green.

PR #1138, opened 2026-09-08
docs(watcher): document the variant suffix and digit precision rule

scttbnsn:
Closes roadmap DR-128 with the rule Scott picked: suffixes must match letter for letter (#1133), digit runs inside a suffix count as precision rather than identity, exact suffix ranks first when both candidates exist, and punctuation-only differences never match (#1136). Users who need a digit to be part of the identity get pointed at `dd.tag.include`.

Docs only: one bullet on the watchers page next to the tag family explanation, a short clause on the `dd.tag.family` label row, and a CHANGELOG Documentation bullet.

### Changelog

✨ Added watcher tag variant rules:
- Suffixes must match letter for letter.
- Digit runs represent precision, not identity.
- Exact suffixes rank first.
- Punctuation-only differences do not match.
- Use `dd.tag.include` when digits must define identity.

🔧 Changed the `dd.tag.family` label documentation to clarify suffix matching and digit-run handling.

### Concerns

- Verify that the documented rules match the v1.8 watcher implementation.


biggest-littlest (approved):
Docs only, matches the DR-128 decision. Gate and CI green.

ALARGECOMPANY (approved):
Docs only, matches the DR-128 decision. Gate and CI green.

PR #1141, opened 2026-09-08
feat(mqtt): register one Home Assistant device per watched container

scttbnsn:
Roadmap 7.8, the devices half promised on #210. Every discovered entity used to sit on a single shared "drydock" device, so a multi-container deployment showed one device with dozens of entities. Each watched container now gets its own device, with the drydock device as `via_device`.

- Device identifier is `drydock_<discovery unique_id>`, so the device id and the entity id come from the same identity key and cannot drift.
- New option `hass.devicepercontainer` (`DD_NOTIFICATION_MQTT_<name>_HASS_DEVICEPERCONTAINER`, default `true`). Set it to `false` to keep the old single-device layout byte for byte.
- No tombstone: the discovery config topic and `unique_id` are unchanged, so entity ids, unique ids and history are preserved. Existing entities move to their container device after an MQTT integration reload or an HA restart, because HA's discovery-update path creates the device but does not move an already-loaded entity. New containers appear under their own device immediately.
- Home Assistant renders discovered entities as `<device name> <entity name>`, which would show every container as `nginx nginx`. With the flag on the update entity's `name` is published as `null`, HA's marker for a device's main feature, so the friendly name becomes `nginx` instead of `drydock nginx`. Entity ids are untouched.
- HA 2024.12 and later stub-creates the parent device on first sight of `via_device`, so publish order does not matter; older HA heals the link on the next discovery publish.

Docs: new option row and a "Devices" subsection with the migration note on the MQTT trigger page. CHANGELOG bullet under Unreleased / Added.

## Changelog

- ✨ Added per-container Home Assistant device registration.
- ✨ Added `hass.devicepercontainer`, defaulting to `true`.
- ✨ Added per-container device identity, model fallback, and `via_device` metadata.
- 🔧 Preserved discovery topics, entity IDs, unique IDs, and history.
- 🔧 Preserved the shared Drydock device for aggregate and watcher entities.
- 🔧 Added migration behavior for existing entities after MQTT reload or Home Assistant restart.
- 🔧 Preserved legacy payloads when `hass.devicepercontainer` is `false`.
- 🔧 Updated MQTT documentation with migration and naming impacts.
- ✨ Added an Unreleased changelog entry.

## Concerns

- Verify automation behavior for entities whose displayed name changes after migration.
- Verify Compose-based device identifiers remain stable when container names change.
- Verify `HASS_DEVICEPERCONTAINER` maps correctly through environment configuration and Joi validation.
- Confirm Home Assistant accepts `name: null` for update entities in all supported versions.


biggest-littlest (approved):
feat(mqtt): register one Home Assistant device per watched container (#1141)

ALARGECOMPANY (approved):
feat(mqtt): register one Home Assistant device per watched container (#1141)

PR #1142, opened 2026-09-08
feat(mqtt): scope plain MQTT state topics per agent behind an opt-in flag

scttbnsn:
Roadmap DR-130, the plain-MQTT facet of #386. With Home Assistant off, the container state topic is `<topic>/<watcher>/<container>` with no agent segment, so two agents that both run a `local` watcher publish the same container name to one topic and overwrite each other.

`DD_NOTIFICATION_MQTT_{trigger_name}_AGENTTOPICSEGMENT=true` inserts `agent/<name>` for agent-owned containers. Off by default so existing subscribers keep the layout they have. Controller-local topics are unchanged either way. Turning it on also turns on `HASS_AGENTTOPICSEGMENT`, because the state payload has to land on the exact topic the HA discovery config names as `state_topic` (the DR-129 bug), and Hass reads the hass-level flag.

The old unscoped retained topic is not swept when the flag flips: a controller-local container of the same name can legitimately own it. Documented on the MQTT page with the manual clear.

Tests: two agents sharing a watcher name land on distinct topics with the flag on and the same topic with it off, controller-local unchanged, the hass flag is forced on, and the schema default is false. CHANGELOG bullet under Unreleased / Added.

## Changelog

- ✨ Added opt-in `agenttopicsegment` for agent-scoped plain MQTT state topics.
- 🔧 Enabled matching `hass.agenttopicsegment` when MQTT agent segmentation is enabled.
- 🔧 Preserved controller-local topic layouts.
- 🔧 Added schema defaults, validation tests, documentation, and an Unreleased changelog entry.

## Concerns

- ⚠️ Existing retained topics remain under the old layout and require manual cleanup.
- ⚠️ Verify provider configuration validation follows the existing mutation and defaulting conventions.


biggest-littlest (approved):
feat(mqtt): scope plain MQTT state topics per agent behind an opt-in flag (#1142)

ALARGECOMPANY (approved):
feat(mqtt): scope plain MQTT state topics per agent behind an opt-in flag (#1142)

PR #1148, opened 2026-09-08
fix(watchers): cap cron deadlines at the supported timer limit

scttbnsn:
Fix the long-cron timer overflow found during images-manager QA. A monthly or longer schedule makes twice the interval exceed Node's timer ceiling. Node then schedules the deadline after 1 ms, clears the running scan guard, and logs a false stall.

Cap the deadline at 2147483647 ms. Keep the existing ten-minute floor and interval-based behavior below that ceiling. Add a real-timer monthly regression and boundary tests. The regressions fail against the old implementation, and all 33 watcher tests pass with the fix.

The full pre-push gate and backend build pass, including app and UI coverage. Add the fix to the not-yet-published rc.11 notes. This holds the maintenance cut until the fix is merged and both exact-source release workflows pass again. Forward ports are prepared for 1.7 and 1.8.

Threat model: ordinary operator cron configuration in a Docker-management service. No hostile local user or new scheduling subsystem is involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Monthly and longer watcher schedules could expire scans after 1 ms.** The scan deadline was twice the cron interval, which overflowed Node's timer limit and cleared the in-flight scan guard almost immediately. Deadlines now stop at the largest supported delay, preserving the existing ten-minute floor and shorter schedule behavior.
+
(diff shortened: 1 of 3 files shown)


scttbnsn:
@coderabbitai review

Review the final cron deadline fix at 27845d58eca18489031de30ce9d1d131a3f236b1. Threat model: an operator-configured Docker update manager; valid long cron schedules must not overflow Node timers or falsely expire the single-flight guard. Hostile operators attacking their own machine are out of scope. Check the timer boundary and regression tests. This is the first actual review; the earlier attempt was rate-limited.

PR #1149, opened 2026-09-08
fix(watchers): forward-port the cron deadline ceiling to v1.8

scttbnsn:
Forward-port the long-cron timer fix from #1148 to dev/v1.8. Monthly and longer intervals could exceed Node's timer limit and expire an in-flight scan after 1 ms. Cap the deadline at 2147483647 ms without changing the existing floor or shorter intervals.

The landed security prerequisite #1152 is now merged into this branch without rewriting history. The current three-dot diff contains only the cron deadline cap, its regression tests, and the changelog entry (3 files, +41/-3). Security guards remain byte-for-byte identical to dev/v1.8.

Validation on fb5813d435657be4680dd0d5a6b3ffe9fd419de1: all 44 focused watcher tests and 16 security guard tests pass. The normal full pre-push gate passed in 343.32 seconds, including scripts/workflow tests, UI typecheck, 100% backend/frontend coverage, and both builds.

Threat model: ordinary operator cron configuration in a Docker-management service. No hostile local user or new scheduling subsystem. Merge behind the 1.6 and 1.7 release work.

## Changelog

- 🔧 Changed cron deadline calculation to cap timers at Node’s `2147483647` ms limit while preserving existing floor behavior.
- ✨ Added monthly schedule and timer-ceiling regression tests.
- 🔒 Pinned Artillery’s `csv-parse` dependency to `7.0.2`.
- 🔧 Added an idempotent postinstall patch for Artillery CSV imports.
- ✨ Added tests for CSV prototype pollution, payload parsing, patch behavior, and dependency resolution.


scttbnsn:
@coderabbitai review

Please review head 91fa92b69911589d451341d230e37bfbc9139eb3. This is review pass 1.

Threat model: a Docker update manager with valid operator-configured long cron schedules. Check that the watch deadline stays within the Node timer range and preserves ordinary schedules, including the regression coverage. A hostile operator attacking their own host is out of scope. The CSV dependency fix and its runtime compatibility tests are included prerequisites. Keep findings scoped to this diff; new scheduling subsystems belong in follow-up work.

scttbnsn:
@coderabbitai review

Please complete pass 2 on exact head fb5813d435657be4680dd0d5a6b3ffe9fd419de1 after merging the landed security prerequisite. The current three-dot diff is only the cron deadline cap, its regression tests, and the changelog entry. The existing security guards match dev/v1.8 exactly.

Threat model: valid operator-configured long schedules in a Docker update manager. A hostile operator attacking their own host is out of scope. No new scheduler or policy subsystem. Check timer ceiling behavior and preservation of ordinary intervals/in-flight guarding. The normal full gate passed in 343.32 seconds, with 44 focused cron tests, 16 security tests, and 100% app/UI coverage.

PR #1189, opened 2026-09-13
fix(ci): require opt-in for second-opinion reviews

scttbnsn:
## Change

Stop CodeRabbit from automatically applying suggested labels. Keep suggestions, but restrict `second-opinion` to high-stakes security-sensitive changes after a completed CodeRabbit source review and explicit maintainer or agent opt-in. Routine UI work and generated charts do not qualify; refactor size or disagreement alone is not a reason to summon another reviewer.

The automatic path was confirmed on #1182: CodeRabbit applied the label at 17:05:44 UTC, triggering Greptile run 34770546941 before CodeRabbit finished its source review at 17:13. This was configuration behavior, not a manual request on that head.

The existing Greptile workflow, pinned reusable workflow, permissions and `skipReview: AUTOMATIC` setting are unchanged. CodeRabbit still reviews every eligible PR, and all required checks and protections remain intact.

## Verification

- Two real configuration regressions failed against the previous policy, then passed after the YAML change.
- All 292 workflow tests across 30 files pass, including the existing label-only trigger and narrow-permissions assertions. Biome, diff checks and normal commit hooks pass.
- Parent review of the complete two-file diff found no actionable issue.

Normal full pre-push passed in 373.92 seconds on `5bb5959195e3683cb26d7f450c48196c7a574a68`: app 16,515 passed plus one existing opt-in skip, UI 5,703 passed, both configured coverage scopes at 100%, actual Vue checker, both builds, 248 script tests, 292 workflow tests, 96 website script tests and applicable static checks. Remote head/tree are exact and the checkout is clean. Hosted CI and the first CodeRabbit source review are required before merge.

This is a review-label policy change only, targeting dev/v1.8. No application runtime, deployment, release or workflow-permission change.

### Changelog

- 🔧 Changed `.coderabbit.yaml` to disable automatic label application.
- 🔒 Restricted `second-opinion` to high-stakes, security-sensitive changes after CodeRabbit source review and explicit maintainer or agent opt-in.
- ✨ Added regression tests for label configuration and scope restrictions.


scttbnsn:
@coderabbitai review

PR #1191, opened 2026-09-13
fix(ui): localize log loading fallback errors

scttbnsn:
## Change

Use the existing translated Logs fallback when an HTTP failure has no usable server diagnostic. Preserve a nonblank server error exactly as returned instead of discarding it. Stream behavior and the shared error helper are unchanged.

## Verification

- Nine real-view regressions failed before the fix: eight unusable error responses and one discarded server diagnostic. They now pass.
- 105 focused tests across the log service, Logs view, related viewers and stream handling pass. Selected log service coverage is 100% in all four metrics. Actual Vue typecheck and Biome pass.
- Parent source review found no actionable issue. Browser QA used the actual Logs view, service and stream adapter with synthetic network boundaries. Pausing live logs and returning an HTML 503 showed the French fallback in wide dark and narrow light layouts. A successful response rendered a real row. Changing the native level filter and returning a JSON diagnostic preserved its exact text.
- The fixture was explicitly remounted between the first error and successful response. The existing toolbar disappears during loading/error; this patch does not add in-view retry behavior. No backend connection or full-app integration is claimed. The owned browser page and listener were stopped and verified absent.

Normal full pre-push passed in 344.48 seconds at `380dfb0cf44c832b547dac9058db2172f23bc83b`. Backend: 16,515 passing tests plus one existing skip; UI: 5,775 passing tests. Both configured coverage gates report 100% in all four metrics. Both builds, actual Vue typecheck, static analysis, 248 script tests, 292 workflow tests and 96 website script tests pass. Remote head and tree match the verified local commit. Hosted CI and the first CodeRabbit source review remain required before merge.

Threat model: HTTP error presentation in the existing authenticated Logs UI. No API, permissions, persistence or stream protocol changes. Targets dev/v1.8, not main or production.

✨ Added
- Added localized fallback handling for log-loading failures without usable server diagnostics.
- Added regression tests for HTTP, network, malformed, and localized failure scenarios.

🔧 Changed
- `getLogEntries` preserves nonblank server `error` values exactly.
- `LogsView` uses `t('logsView.loadFailed')` only when `errorMessage` returns an empty value.
- Stream behavior and shared error handling remain unchanged.

- Verify that JSON response parsing does not consume a response body required elsewhere.
- Keep the existing `errorMessage` helper as the single source for error normalization.


scttbnsn:
Integrated the landed Vue coverage gate at `64f38bc9598bef49f8277a5da4de1085c14241c5`. The four Logs source/test files are unchanged. Full pre-push passed in 339.28s: 16,515 backend tests plus one skip, 5,797 UI tests, backend and all 154 UI TypeScript files at 100%, and all 100 Vue components above the unchanged floors. Both builds and the Vue checker passed. Remote head matches; checkout and workers are clean.

Hosted checks are running. CodeRabbit's automatic response is a quota notice, not a source review, so this stays unmerged until that review is available. No second-opinion request for this localized error-message fix.

scttbnsn:
@coderabbitai review

Please review exact head 64f38bc9598bef49f8277a5da4de1085c14241c5 now that the included review window has reset. This is the first source review; earlier responses were quota notices. Scope: localized log-fetch error presentation, retained data, locale changes and existing request behavior. No authentication or authorization changes.

PR #1196, opened 2026-09-14
fix(ui): retry failed paused log requests

scttbnsn:
## Summary

Paused application logs can get stuck behind a failed fetch because the error state replaces the viewer and its controls. Add a translated Retry button so the same mounted page can recover.

- Retry reuses the current filters and existing fetch path, preserves server diagnostics, and leaves the stream paused.
- Ignore retry while live or already loading. Existing filter requests and explicit Resume behavior stay unchanged.
- Reuse `common.retry` in all 17 locale catalogs. No API, backend, subscription, or dependency changes.

## Verification

- Real mounted HTTP/WebSocket regression first failed in three recovery cases because no accessible Retry button existed; the live-mode control case passed. All four cases pass after the fix.
- Integrated focused checks: 87 tests across eight files, plus the actual Vue typecheck and UI build.
- Full normal pre-push gate passed in 363.67 seconds on `0cd6ebfe65858fc6a564a3e7dec8deaf60386036`: 16,582 backend tests passed (one existing skip), 5,911 UI tests passed, 250 script tests, 292 workflow tests, 96 website-script tests, lint/static checks, Vue typecheck, both builds, and workflow security checks. No bypasses or coverage-floor changes.
- Backend and all 156 UI TypeScript files retain 100% statement/branch/function/line coverage. All 101 Vue components measured 88.44% statements, 82.77% branches, 85.84% functions, and 88.83% lines, above the existing component floors.
- Focused coverage for the two affected components: 93.79% statements, 92.59% branches, 89.09% functions, and 94.24% lines. New retry logic and guards are covered; existing coverage floors are unchanged.
- Browser QA on the integrated head used the real components and services with synthetic HTTP/WebSocket boundaries. Mouse retry recovered 503 to rows without remounting; keyboard Enter recovered a filtered 503 to an empty result while retaining `warn`. Explicit Resume made no extra HTTP request, and unmount closed the stream. One mount, ten GETs, four entries requests, both sockets closed, zero boundary violations, and no French-layout overflow at 393px.

Scope: five files, limited to the Logs view/component, their tests, and changelog. Browser QA is component/service integration, not a full backend or Docker test.


scttbnsn:
@coderabbitai review

The last attempt hit the review limit without producing a review. Please review head 0cd6ebfe65858fc6a564a3e7dec8deaf60386036. Scope: retrying failed paused-log requests in the existing authenticated Docker management UI. Check retry, loading and pause state transitions; no new auth or multitenant model is introduced.

PR #1204, opened 2026-09-15
fix(mqtt): align Home Assistant discovery and state topics

scttbnsn:
Backports the shared MQTT topic construction from #1139 to the 1.7 line for discussion #1201. Discovery and state publication now use the same canonical container identity and optional agent segment. The existing 1.7 default stays enabled, and plain MQTT topics are unchanged when Home Assistant discovery is disabled.

Credit to Scott Benson for the original implementation in #1139.

Validation: regression cases failed before the fix; focused MQTT tests pass with 100% coverage. The full pre-push gate passed with 100% backend and UI coverage, builds, lint, static analysis, and workflow tests. All required hosted checks passed on c9cb4a6c977bcaeb5268f90c03659ebc1ca557b5, and CodeRabbit found no actionable issues. The reporter confirmed 1.7.0-rc.14 and the agent-scoped discovery topic, matching the mismatch in that tag’s publisher. Live broker verification on the fixed build remains outstanding. The accompanying Alpine timezone pin repair also passed both hosted image builds.

PR #1208, opened 2026-09-15
fix(tests): accept safe dependency updates in version guards

scttbnsn:
Fix version guards that reject valid dependency updates in #1199 and #1200. Require exact stable versions and matching installed lockfile versions instead of freezing the tests to the package version that happened to be current when written.

The YAML security minimum stays at 2.9.0, and every nested YAML install must match the exact override. Missing entries, older versions, ranges, prereleases and mismatches fail. PostHog must match the manifest, lock root and installed package; removed Vercel telemetry packages still fail. Analytics privacy behavior and its tests are unchanged.

Reproduced the failures against both Renovate PR manifests and lockfiles. The focused runs pass 65 UI security tests and 143 website-script tests, including invalid-fixture coverage. The full normal pre-push gate passed in 307.14 seconds with 100% app/UI coverage, both builds, lint, Knip, Qlty and workflow tests. No dependency version or bot history is changed by this PR. Clean Docker builds also need the timezone-pin fix in #1204; that maintenance fix remains separate from this test-only patch.


scttbnsn:
@coderabbitai review

Please review exact head 5fe6af2ee613b4cf3f2601cc7e781953531cc705. This makes dependency guards accept newer exact stable pins while keeping the YAML security floor and nested-install agreement, and PostHog manifest/root/install equality. It includes the minimum PostHog 1.427.2 correction you verified in #1209, with the same guard file bytes. Invalid versions, downgrades, missing entries and Vercel telemetry packages remain rejected; privacy assertions are unchanged. The full local gate passed in 306.42 seconds with 100% app/UI coverage and 150 website scripts. Threat model: ordinary dependency/lockfile drift, not a hostile maintainer rewriting tests. This is the first review request for this PR.

PR #1253, opened 2026-09-18
ci: update v1.6 runner control-plane support

scttbnsn:
## Changes

Update the v1.6 Harden Runner pin from 2.20.0 to 2.21.1 (`e14015d583714f6e62063499dc959a02595150a1`). This carries only the runner update already present on v1.7 through #1197, not the rest of that dependency batch.

The upstream maintainer identifies missing GitHub runner endpoints as the cause of block-mode runner shutdowns in [step-security/harden-runner#679](https://github.com/step-security/harden-runner/issues/679#issuecomment-5471947960). The [2.21.1 release](https://github.com/step-security/harden-runner/releases/tag/v2.21.1) fixes that allowance. Its agent update loads GitHub Actions domains into the implicit endpoint list.

Drydock's failed coverage jobs have the same shutdown signature. The separate #1250 Grype job also logged blocked GitHub watchdog/orchestrator endpoints. This is a reason to apply the upstream fix, not proof that every Drydock failure has the same cause.

All 24 local runner uses remain SHA-pinned. Egress policies, explicit endpoint lists, sudo restrictions, permissions, job conditions and timeouts are unchanged. Reusable external workflows and product source are untouched. The zlib image finding is separate and remains unsuppressed.

## Verification

- Updated the existing workflow pin expectations first: two failures identifying all 24 stale jobs, three policy-preservation checks passed.
- Applied the new workflow pins: all five focused checks passed.
- Compared all eight changed workflow files against their originals after normalizing only the SHA/version comment: all other bytes are identical.
- Full normal pre-push passed in 346.18 seconds, including 100% configured backend/UI coverage, both builds, all workflow tests, static checks and Zizmor. No bypass or relaxed gate. Hosted execution must verify the fix; no reliability claim yet.

## Changelog

- 🔒 Upgrade all 24 SHA-pinned `step-security/harden-runner` uses from `v2.20.0` to `v2.21.1`.
- 🐛 Apply upstream fixes for missing GitHub Actions runner endpoints during block-mode shutdowns.
- 🔧 Update test expectations to commit `e14015d583714f6e62063499dc959a02595150a1`.
- 🔧 Preserve existing policies, permissions, conditions, timeouts, reusable workflows, sudo restrictions, and product source.

## Concerns

- Hosted execution is required to verify the runner endpoint fix.
- The zlib image finding remains unsuppressed.


biggest-littlest (approved):
Verified upstream v2.21.1 pin and clear CodeRabbit review 84c3ca93 at this exact head. All workflow changes are pin/comment-only; policy and permission settings are preserved. Full local gate passed. Hosted checks, including the separate image vulnerability gate, must pass before merge.

PR #1264, opened 2026-09-18
fix(ui): show the active SQLite store filename

scttbnsn:
General settings showed `dd.json` even when the active store was SQLite. The API returns that legacy import filename alongside `dbFile`; the UI was reading the wrong field.

This uses the configured database filename, including custom names, and keeps the old response fallback when `dbFile` is absent. No storage or migration behavior changes.

Verification: reproduced against a running backend with a fresh SQLite store. Both new mounted-component regressions failed before the fix and passed afterward. All 46 targeted tests and Vue typechecking pass. The full normal pre-push gate passed in 355.42 seconds, including configured app/UI coverage and both builds.

Post-fix browser check also passed on a local-only combined v1.8 rehearsal at 4496339070e7c6f21f844d877dbf0e409784b34a. The real API returned both legacy file=dd.json and custom dbFile=qa-storage.sqlite; General settings displayed qa-storage.sqlite. The three source/test files match this PR exactly. Both combined builds and all 222 focused tests passed. This is not Docker, mobile-layout or full-release QA.

## Changelog

- ✨ Added optional `dbFile` to `StoreConfiguration`.
- 🔧 Changed General settings to display `dbFile`, then legacy `file`, then the localized unknown value.
- 🐛 Fixed display of custom SQLite filenames instead of always showing `dd.json`.
- 🔧 Added coverage for SQLite filenames and the legacy fallback.

## Concerns

- None identified.


biggest-littlest (approved):
Checked the filename selection and mounted-component regressions. CodeRabbit completed its source review without actionable findings. This is display-only; storage and migration behavior are unchanged.

PR #1265, opened 2026-09-18
fix(ui): omit unconfigured dashboard hosts

scttbnsn:
The dashboard invented a connected Local host at `/var/run/docker.sock` whenever there were no non-agent watchers. That also affected agent-only deployments with the local watcher disabled.

Build host rows only from configured watchers and agents, matching the Hosts page. Existing host addresses, connection states, container counts and grouping complexity stay unchanged. Stale local container inventory does not create a host.

Four new regression cases failed before the fix. Tests also cover the mounted dashboard, configured Local rows and agent-only fleets. All 352 dashboard/host tests and TypeScript checking pass using this release line's locked dependencies. The full normal pre-push gate passed in 354.49 seconds, including configured app/UI coverage and both builds. Docker and post-fix browser verification are not claimed.

Draft while earlier PRs use the limited source-review slots. No review or release gate is bypassed.

## Changelog

- 🐛 Fixed dashboard host generation to use only configured non-agent watchers.
- 🗑️ Removed the fallback `Local` host at `/var/run/docker.sock`.
- ✨ Added coverage for configured Local hosts, empty watcher configurations, and agent-only fleets.
- 🔧 Preserved host addresses, connection states, container counts, and grouping behavior.

## Validation

- 352 dashboard/host tests passed.
- TypeScript checking passed.
- Full pre-push gate passed.

## Concerns

- Docker and post-fix browser verification were not performed.


biggest-littlest (approved):
Checked the exact-head diff and completed CodeRabbit review. Configured hosts remain intact; empty and agent-only inventories no longer invent a Local host. Existing full-gate evidence covers this head. Image security must still pass before merge.

PR #1269, opened 2026-09-19
fix(ui): translate shared display and event captions

scttbnsn:
## Fix

Fill English copies of shared yes/no and unnamed custom-registry labels, plus missing-profile and paused/offline log captions, in 13 language catalogs. Also translate maintenance-window states, maturity-cleared audit captions and the synthetic test-container name across all 16 non-English locales. Spanish maintenance wording now describes the current state rather than an action. Submitted booleans, real registry/user identities, raw log messages, schedules and streaming state stay unchanged.

The Arabic Custom translation matches the same pending leaf in #1258. That PR still supplies the ContainersView translator binding; this catalog patch does not replace it.

## Verification

- Original shared-label change: fifteen new tests failed before the catalog fix; four already-translated locale cases passed. All 17 locale cases then passed with the real display helper. GHCR, registry hostnames and configured registry names stay unchanged.
- The real notification editor updates mounted German/Arabic boolean labels without refetching, preserves the selection, and submits boolean false through the existing service.
- Configuration follow-up: 26 tests reproduced 39 untranslated profile/log captions before the catalog fix. Tests use the real ConfigView and ConfigLogsTab/AppLogViewer, update the locale without remounting, preserve a real username matching the English fallback and raw log text, and check that locale changes do not reload profiles, write settings, toggle streaming or retry.
- State/event follow-up: 48 new locale cases failed before the catalog changes. Tests retain the original watcher data and selected audit entry, preserve server diagnostics, and verify that only an explicit test-button click submits the localized synthetic name, with trigger/agent/id/image/update-kind unchanged.
- Latest focused run: 336 tests across seven suites passed, plus Vue typechecking, Biome and diff checks.
- Normal pre-push gate passed for 4ff0c71e296c8a96e4905a9404c3f58d0c37ebcf in 354.63 seconds: configured backend/UI coverage, both builds, 284 script tests, 318 workflow tests, Vue typechecking and static checks. Website scripts, local Docker and workflow-security scanning correctly skipped under their normal conditions. No browser or screenshot claim. Hosted checks and source review remain separate gates.

Review scope: display catalogs and regression tests only. No changes to authentication, action permissions, notification routing, configuration ownership or API schemas. The required image-security gate remains a separate blocker; it has not been bypassed.


biggest-littlest (approved):
CodeRabbit completed the full source review at this exact head with no actionable comments. The local full gate passed; hosted image-security checks still apply.

PR #1272, opened 2026-09-19
fix(registry): bound Docker Hub metadata requests

scttbnsn:
## Fix

Docker Hub publish-date metadata requests omitted the shared outbound HTTP timeout. A stalled response could keep container discovery waiting before its first inventory report. Apply `DD_OUTBOUND_HTTP_TIMEOUT_MS`, including its existing default and invalid-value fallback, to that request.

Authentication, TLS, redirect restrictions, rate limiting and bounded retries stay unchanged. The watcher already handles exhausted metadata failures without losing its image comparison result.

## Verification

- Five regression assertions failed before the fix, including a real Axios request to a delayed loopback HTTP response that returned late instead of timing out.
- After the fix, 159 tests passed across the Hub provider, timeout configuration, HTTP retry helper, image comparison and container processing.
- The loopback server, sockets, timer and environment overrides are cleaned up in `finally`.
- Full normal pre-push passed in 350.76 seconds: configured app/UI coverage, both builds, UI typecheck, static checks, 197 script tests, 102 workflow tests, 57 website script tests and Zizmor.

Found while investigating the Cucumber inventory-readiness failure on #1269. This fixes a reproduced defect, but the available hosted logs do not prove it caused that particular failure. No CI timeout increase or fixture exclusion.

This is the oldest affected release line. Forward ports to 1.7 and 1.8 follow separately.

✨ Added
- Added Docker Hub timeout coverage for configured, default, invalid, and stalled responses.

🔧 Changed
- Applied `getRegistryRequestTimeoutMs()` to Docker Hub metadata requests.
- Reused the existing registry timeout utility and `DD_OUTBOUND_HTTP_TIMEOUT_MS` configuration.

🐛 Fixed
- Stalled Docker Hub metadata requests now abort at the configured timeout.
- Unset or invalid timeout values use the 30,000 ms default.

Concerns:
- No actionable issues identified from the supplied evidence.
- Review finding severity counts are unavailable.


biggest-littlest (approved):
Reviewed the metadata-only timeout change and its regression coverage. CodeRabbit completed its first source review with no findings. The full local gate passed; required CI still applies.

PR #1273, opened 2026-09-19
fix(registry): bound Docker Hub metadata requests on v1.7

scttbnsn:
## Fix

Forward-port #1272 onto `dev/v1.7`. Docker Hub publish-date metadata requests now use the configured outbound HTTP timeout, so a stalled response cannot indefinitely hold up container discovery.

The provider and regression test files are byte-identical to the 1.6 fix. Authentication, TLS, redirect restrictions, rate limiting, bounded retries and publish-date failure handling stay unchanged.

## Verification

- Original test-first proof in #1272: five assertion failures before the fix, including a real delayed HTTP response resolving instead of timing out.
- All 159 focused tests passed on the 1.7 dependency tree across the Hub provider, configuration, HTTP retry helper, image comparison and container processing.
- Biome and diff checks passed.
- Full normal pre-push passed in 321.99 seconds: configured app/UI coverage, both builds, UI typecheck, static checks, script/workflow/website tests and Zizmor.

Keep this behind the 1.6 fix in release order. This addresses a reproduced missing timeout, not a proven root cause of the historical Cucumber failure on #1269.

## Changelog

- 🐛 Fixed Docker Hub metadata requests to use `getRegistryRequestTimeoutMs()`.
- ✨ Added tests for default, numeric, and invalid timeout configuration.
- ✨ Added a stalled-response test that verifies `ECONNABORTED`.

## Concerns

- Review findings: unavailable.
- Test and verification status: unavailable from the supplied change summary.


biggest-littlest (approved):
Reviewed the exact forward port and regression tests. CodeRabbit completed its first source review with no findings, and the full local gate passed. Keep the 1.6 fix ahead of this port.

PR #1274, opened 2026-09-19
fix(registry): bound Docker Hub metadata requests on v1.8

scttbnsn:
## Fix

Forward-port #1272 and #1273 onto `dev/v1.8`. Docker Hub publish-date metadata requests now use the configured outbound HTTP timeout, so a stalled response cannot indefinitely hold up container discovery.

The provider and regression test files are byte-identical to the 1.6 fix. Only the changelog insertion needed adaptation. Authentication, TLS, redirect restrictions, rate limiting, bounded retries and publish-date failure handling stay unchanged.

## Verification

- Original test-first proof in #1272: five assertion failures before the fix, including a real delayed HTTP response resolving instead of timing out.
- All 161 focused tests passed on the 1.8 dependency tree across the Hub provider, configuration, HTTP retry helper, image comparison and container processing.
- Biome and diff checks passed.
- Full normal pre-push passed in 367.33 seconds: configured app/UI coverage, both builds, UI typecheck, static checks, script/workflow/website tests and Zizmor.

Keep this behind the 1.6 and 1.7 fixes in release order. This addresses a reproduced missing timeout, not a proven root cause of the historical Cucumber failure on #1269. The subsequent #1271 Cucumber run passed without this patch.

## Changelog

- 🐛 Fixed Docker Hub metadata requests to use `getRegistryRequestTimeoutMs()`.
- ✨ Added regression tests for configured, default, invalid, and stalled request timeouts.
- 🔧 Preserved existing authentication, TLS, redirect, rate-limit, retry, and failure-handling behavior.
- ✅ Verification passed: 161 focused tests, Biome checks, diff checks, and the full pre-push suite.

## Concerns

- Confirm release sequencing with the corresponding 1.6 and 1.7 fixes.


biggest-littlest (approved):
The metadata request now uses the shared registry timeout. Checked the focused diff and real-HTTP timeout regression. CodeRabbit completed its source review without actionable findings; image-security remediation is still required before merge.

